PR #24, opened 2025-02-04
Nhse o34 orkv.i61 maybedeferfetch

martinsumner:
Uses https://github.com/OpenRiak/riak_core/pull/14 to improve efficiency of repairs with the leveled backend

This allows the deferring of fetches during a handoff fold.  If the heads capability exists it will now return a stream of "proxy objects" not objects, and only if it has been decided to send that object (i.e. it has passed all the FilterFuns) will the actual object be fetched.

One could extend this further in the future i.e. check for non-presence of the object in the handoff target, before fetching the object.

As part of the change, if there are issues decoding an object a read-repair is prompted - that should resolve any issue in both sender and receiver.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -2467,7 +2467,7 @@ handoff_started(SrcPartition, WorkerPid) ->
                         [
                             {
                                 repair,
-                                [{fold_heads, true}, {check_presence, true}]
+                                [{fold_heads, true}, {check_presence, defer}]
                             }
                         ];
                     _ ->
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -95,6 +95,7 @@
 -include_lib("riak_kv_map_phase.hrl").
 -include_lib("riak_core_pb.hrl").
 -include("riak_kv_types.hrl").
+-include("riak_kv_capability.hrl").
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -2488,9 +2489,8 @@ encode_handoff_item({B, K}, V) ->
     %% before sending data to another node change binary version
     %% to one supported by the cluster. This way we don't send
     %% unsupported formats to old nodes
-    ObjFmt = riak_core_capability:get({riak_kv, object_format}, v0),
     try
-        Value  = riak_object:to_binary_version(ObjFmt, B, K, V),
+        Value  = riak_object:to_binary_version(?CAP_OBJECT_FORMAT, B, K, V),
         encode_binary_object(B, K, Value)
     catch Error:Reason ->
             ?LOG_WARNING("Handoff encode failed: ~p:~p",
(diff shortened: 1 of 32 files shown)


src/riak_kv_vnode.erl
```diff
@@ -2519,8 +2535,9 @@ encode_handoff_item({B, K}, V) ->
         Value  = riak_object:to_binary_version(?CAP_OBJECT_FORMAT, B, K, V),
         encode_binary_object(B, K, Value)
     catch Error:Reason ->
-            ?LOG_WARNING("Handoff encode failed: ~p:~p",
-                          [Error,Reason]),
+            ?LOG_WARNING(
+                "Handoff encode failed: ~0p:~0p", [Error, Reason]),
+            riak_kv_reader:request_read({B, K}),
```
ThomasArts: would be nice with a little comment on why this read request is needed here. Not obvious from the code.

src/riak_kv_vnode.erl
```diff
@@ -3674,15 +3691,15 @@ do_fold(Fun, Acc0, Sender, ReqOpts, State=#state{async_folding=AsyncFolding,
 %% then the fold_heads function can be used on the backend if it suppports that
 %% capability.
 maybe_use_fold_heads(Capabilities, Opts, Mod) ->
-    case lists:member(fold_heads, Opts) of
+    case proplists:get_value(fold_heads, Opts) of
```
ThomasArts:
```suggestion
    case proplists:get_value(fold_heads, Opts) andalso lists:member(fold_heads, Capabilities) of
```
and then just one level deep case, true fold heads, false fold objects

Another nice function is `proplists:get_bool(fold_heads, Opts)`

src/riak_object.erl
```diff
@@ -1197,8 +1197,13 @@ to_binary_version(v0, _, _, <<131,_/binary>>=Bin) ->
     Bin;
 to_binary_version(v1, _, _, <<?MAGIC:8/integer, 1:8/integer, _/binary>>=Bin) ->
     Bin;
-to_binary_version(Vsn, B, K, Bin) when is_binary(Bin) ->
-    to_binary(Vsn, from_binary(B, K, Bin));
+to_binary_version(v1, B, K, Bin) when is_binary(Bin) ->
```
ThomasArts: This code changes from being able to handle all binaries (left to match in this cluase) to only binaries for `v1`. 
That is, if there had been a `v0` or there will be a `v2`, then this code would fail for binaries. Not sure that's a problem, just observe incompatibility

ThomasArts (approved):

src/riak_object.erl
martinsumner: The object format is still configurable - so could be v0.  This would only be required when going back from v1 to v0.  I don't know if this is supported - but preventing it here is unnecessary so will fix to allow both v0/v1.

Will only get a proxy object if it is v1 (as this is checked before requesting handoff), which is perhaps why v1 was specified ... but downgrade with a non-proxy object is a different case

PR #25, opened 2025-02-05
Add support for direct _fsm as well as direct_stat

martinsumner:
The "direct_" configuration switches will bypass sidejob for stat updates and GET/PUT FSMs.  This should increase throughput in CPU-bound environments, but stops the rate-limiting of sidejob.

Creates dummy stats to replace sidejob stats.

### DIFF
--- a/src/riak_kv_status.erl
+++ b/src/riak_kv_status.erl
@@ -83,13 +83,38 @@ are_indexes_fixed(riak_kv_multi_backend, {_Idx, [{backend_status,_,Status}]}) ->
     fixed_index_status(riak_kv_eleveldb_backend, Statuses).
 
 get_stats(web) ->
-    aliases()
-        ++ expand_disk_stats(riak_kv_stat_bc:disk_stats())
-        ++ riak_kv_stat_bc:app_stats();
+    expand_disk_stats(riak_kv_stat_bc:disk_stats())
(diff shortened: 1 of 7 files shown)


src/riak_kv_status.erl
```diff
@@ -83,13 +83,38 @@ are_indexes_fixed(riak_kv_multi_backend, {_Idx, [{backend_status,_,Status}]}) ->
...
-    aliases()
-        ++ riak_kv_stat_bc:disk_stats()
-        ++ riak_kv_stat_bc:app_stats().
+    riak_kv_stat_bc:disk_stats()
+    ++ riak_kv_stat_bc:app_stats()
+    ++ fake_sidejob_stats()
+    ++ aliases().
+
+fake_sidejob_stats() ->
+    case application:get_env(riak_kv, direct_fsm, false) of
```
WarpEngineer, line 97: if `direct_fsm` is `on` but `direct_stat` is `off` this will end up trying to return `sidejob` stats, right? Is that the intention?

martinsumner:
I'm going to pull this PR until the work on CRDT counters performance is complete.  There's a lot of spinning plates at the moment for both stats and FSM initialisation, and so this probably needs to be revisited in a few weeks when that is all settled.

PR #26, opened 2025-02-06
Nhse 034 orkv.i57 fixcapability

martinsumner:
Capabilities are used to determine if a feature is safe to use, where this is determined by all nodes in the cluster being able to use that capability.

Capability calls are spread around the code.  Some capabilities continue to be check, even though there is no support for a version sufficiently old to not support the capability.

There is a minimal cost to capability checks, but that cost is non-zero.  There are also potential failure scenarios where there may be a fallback to the "wrong" capability.

Unify capability definitions, and where they are frequently requested and backwards-compatibility is no longer possible, hardcode them.

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -70,6 +70,7 @@
          handle_corrupted_object/4]).
 
 -include_lib("kernel/include/logger.hrl").
+-include("riak_kv_capability.hrl").
 
 -type index() :: non_neg_integer().
 -type index_n() :: {index(), non_neg_integer()}.
@@ -547,7 +548,7 @@ check_upgrade_env() ->
 
 -spec get_cap_hash_version() -> version().
 get_cap_hash_version() ->
-    riak_core_capability:get({riak_kv, object_hash_version}, legacy).
+    ?CAP_OBJECT_HASH_VERSION.
 
 -spec find_version(list(), index()) -> version().
 find_version(Root, Index) ->
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -561,9 +561,11 @@ check_root_version(Root, Index, Version) when is_integer(Version) ->
             Version;
         false ->
             legacy
-    end;
-check_root_version(_Root, _Index, Version) ->
-    Version.
+    end
+%     ;
+% check_root_version(_Root, _Index, Version) ->
+%     Version
+    .
 
 %% @doc Init the trees.
 %%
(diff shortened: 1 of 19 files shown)


src/riak_kv_put_fsm.erl
```diff
@@ -189,17 +190,16 @@ get_put_coordinator_failure_timeout() ->
 
 make_ack_options(Options) ->
     AckOption = get_option(ack_execute, Options),
-    AckCap = riak_core_capability:get({riak_kv, put_fsm_ack_execute}, disabled),
     RetryCoord =
         app_helper:get_env(riak_kv, retry_put_coordinator_failure, true) andalso
         get_option(retry_put_coordinator_failure, Options, true),
-    case {AckOption, AckCap, RetryCoord} of
+    case {AckOption, ?CAP_PUTFSM_ACK, RetryCoord} of
```
ThomasArts: When this now always is enabled, this second argument of the tuple can be left out completely. This simplifies the code quite a bit:

```
case RetryCoord andalso not is_pid(AckOption) of
  true -> {true, [{ack_execute, self()}| Options]};
  false > {false, Options}
```

src/riak_kv_index_hashtree.erl
```diff
@@ -560,9 +561,11 @@ check_root_version(Root, Index, Version) when is_integer(Version) ->
             Version;
         false ->
             legacy
-    end;
-check_root_version(_Root, _Index, Version) ->
-    Version.
+    end
+%     ;
+% check_root_version(_Root, _Index, Version) ->
```
ThomasArts: Weird, you should get a type error here. Since `legacy` is a valid type for `version()` as output, it should also be valid as input and then the first clause won't match. So, that would be detectable with a type checker, one had hoped...

ThomasArts (approved):
It might be a good thing to keep the old defaults as comment in the `riak_kv_capability.hrl` header file. On the other hand, newcomers might be confused by seeing that some capabilities have been configurable before but are fixed now.

Should you also remove schema definitions for those capabilities that you cannot set any more?

src/riak_kv_put_fsm.erl
martinsumner: I've retained the history of the capability, but also simplified the case statement

src/riak_kv_index_hashtree.erl
martinsumner: Clarified types

martinsumner:
> It might be a good thing to keep the old defaults as comment in the `riak_kv_capability.hrl` header file. On the other hand, newcomers might be confused by seeing that some capabilities have been configurable before but are fixed now.
> 
> Should you also remove schema definitions for those capabilities that you cannot set any more?

Object format is the only configurable one, and that remains configurable

PR #30, opened 2025-02-27
Nhse o32 orkv.i29 leveledasync

martinsumner:
Allow use of {F, async} response to data_size/1 backend function.

The function dialyzer specs within the backend callback have been extended to at least recognise arity.

### DIFF
--- a/src/riak_kv_leveled_backend.erl
+++ b/src/riak_kv_leveled_backend.erl
@@ -616,9 +616,7 @@ status(_State) ->
 
 %% @doc Get an estimate of the data_size for this leveled backend
 -spec data_size(state()) ->
-    undefined |
-    {non_neg_integer(), objects} |
-    {fun(() -> {non_neg_integer(), objects}), dynamic}.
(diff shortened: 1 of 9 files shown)


martinsumner:
Depend son https://github.com/OpenRiak/riak_core/pull/19

src/riak_kv_leveled_backend.erl
```diff
@@ -616,9 +616,7 @@ status(_State) ->
 
 %% @doc Get an estimate of the data_size for this leveled backend
 -spec data_size(state()) ->
-    undefined |
-    {non_neg_integer(), objects} |
-    {fun(() -> {non_neg_integer(), objects}), dynamic}.
+    {fun(() -> {non_neg_integer(), objects}), async}.
```
ThomasArts, line 619: This is not the same type as used in `riak_kv_backend`. It is a proper subtype, so that should be fine.

ThomasArts (approved):

PR #35, opened 2025-03-20
Don't restart backend on drop

martinsumner:
Hard to understand why this was ever this way.  No other backend restarts after a drop.  Has existed since initial commit of riak_kv_leveled_backend.

Creates confusion for the operator looking at logs.


ThomasArts (approved):
This fix is correct.

PR #43, opened 2025-07-11
Nhse o34 orkv.i42 aae tree exclude

martinsumner:
Add support for aae_tree_exclude feature - https://github.com/orgs/OpenRiak/discussions/14

### DIFF
--- a/docs/BucketProperties.md
+++ b/docs/BucketProperties.md
@@ -1,0 +1,131 @@
+# Bucket Properties
+
+A summary of the bucket properties that can be configured in Riak.
+
+There are two types of Buckets in Riak - non-typed buckets, and typed buckets.  Typed buckets were introduced to make it easier to expand the number of buckets that can be supported with non-default properties.  
+
+A number of "defaults" for bucket properties are configurable via `riak.conf` e.g.
+
+```
+buckets.default.n_val = 3
+buckets.default.merge_strategy = 2
+buckets.default.pw = 1
+buckets.default.allow_mult = true
+...
+```
+
+Configuring these defaults will impact only non-typed buckets.  So any bucket name used where no type is specified will inherit these defaults, but any typed bucket created will NOT inherit these configured defaults - typed buckets instead have fixed, pre-defined defaults.
+
+Two pre-defined defaults changed with the introduction of typed buckets (the merge strategy aka `dvv_enabled`, and the `allow_mult` configuration), it is strongly recommended to configure your clusters to have the new default properties for non-typed buckets to avoid confusion with non-typed buckets having different defaults i.e. by adding to your `riak.conf`:
+
+```
+buckets.default.merge_strategy = 2
+buckets.default.allow_mult = true
+```
+
+As any change made to `buckets.default.*` configuration in `riak.conf` is not inherited for typed buckets, there is no way of changing the defaults for typed buckets, so the operator is required to ensure that all default properties are manually set on every type.  For example if you wish to change the default n_val to 5 - this needs to be changed in riak.conf `buckets.default.n_val = 5` but ALSO the property `{n_val, 5}` has to be added on every single bucket type created.
+
+In general, setting bespoke bucket properties should be done using typed buckets due to the relative efficiency of the implementation with types, but changes to defaults should be considered very carefully.  Bespoke properties allow for bespoke behaviours, but bespoke behaviours add to the cognitive load of future operators.
+
+Default changes made via riak.conf need to be set consistently across a cluster.  No bucket properties are gossipped between clusters, so properties are cluster-specific.  In general any cluster setting related to vector clocks MUST be configured consistently across replicating clusters e.g. `dvv_enabled`, `old_vclock`, `young_vclock`, `big_vclock` and `small_vclock`.  Other properties can be different between clusters. 
+
+Some changes can be applied using GET/PUT specific parameters, which will override the default bucket property i.e. a bucket could be configured to use `{sync_on_write, one}` but a specific PUT can override this by setting `{sync_on_write, all}`.  Although the use of GET/PUT specific parameters is supported, it is is not recommended.  Operation-specific parameters that override defaults are not logged, and can considerably increase the operator challenges when troubleshooting intermittent problems.
+
+## dvv_enabled
+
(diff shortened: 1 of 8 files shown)


martinsumner:
Linked pull requests:

https://github.com/OpenRiak/riak-erlang-client/pull/2
https://github.com/OpenRiak/riak_pb/pull/3
https://github.com/OpenRiak/riak_core/pull/24
https://github.com/martinsumner/kv_index_tictactree/pull/131

martinsumner:
Tests: https://github.com/OpenRiak/riak_test/pull/9

src/riak_kv_pb_object.erl
```diff
@@ -369,6 +394,17 @@ process(
             DW = decode_quorum(DW0),
             PW = decode_quorum(PW0),
             NodeConfirms = decode_quorum(NodeConfirms0),
+            SyncOnWrite =
```
ThomasArts: Should this not be done by a decode function in `riak_pb`? Feels a bit weird to see the decoding happen here.

docs/BucketProperties.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `sync_on_write` property is used only for PUTs via the API.  Internal PUTs (e.g. for transfers) will ignore the property and use the backend configuration, and so will not carry the overhead of flushing.
+
+It is recommended not to use backend sync configuration, and instead control flushing only through use of this bucket property.
+
+If replicating between clusters and `one` is used as the `sync_on_write` bucket property, then the cluster that receives the PUT form the application will flush to disk on one node - but all clusters receiving the PUT via replication will not be required to flush to disk on any node.  The properties of `backend` or `all` are treated equally in source and sink clusters.
```
ThomasArts:
```suggestion
If replicating between clusters and `one` is used as the `sync_on_write` bucket property, then the cluster that receives the PUT from the application will flush to disk on one node - but all clusters receiving the PUT via replication will not be required to flush to disk on any node.  The properties of `backend` or `all` are treated equally in source and sink clusters.
```

docs/BucketProperties.md
```diff
@@ -0,0 +1,139 @@
...
+
+## allow_mult
+
+The `allow_mult` bucket property has a default value of `true`, for any typed bucket, but a default value of `false` for any untyped bucket.  It is recommended to use the value of `true`.
+
+The internal workings of Riak are identical for the two allow_mult settings, with the exception of the case when an unresolvable conflict is discovered in the object change history.  In this case: if `{allow_mult, true}`, all conflicting versions are returned to the client to resolve (on the next GET); if `{allow_mult, false}` only the object with the most recent last_modified_date is returned.
+
+The last_modified _date is a microsecond-level timestamp, that depends on the accuracy of the local node's clock.  If the timestamps match of conflicting changes, then an arbitrary choice is made, although there is a preference for changes with values over deletions.  When using `{allow_mult, false}`, the use of reliable time sources to co-ordinate time within and across clusters is strongly recommended.
```
ThomasArts:
```suggestion
The last_modified_date is a microsecond-level timestamp, that depends on the accuracy of the local node's clock.  If the timestamps match of conflicting changes, then an arbitrary choice is made, although there is a preference for changes with values over deletions.  When using `{allow_mult, false}`, the use of reliable time sources to co-ordinate time within and across clusters is strongly recommended.   
```

docs/BucketProperties.md
```diff
@@ -0,0 +1,139 @@
...
+
+Changing an n_val on a bucket which already contains data will have unexpected and untested consequences, especially when contracting the n_val.
+
+## node_confirms
+
+The `node_confirms` bucket property has a default value of `0`, and may be set to any non-negative integer less than or equal to the n_val (for that bucket).  The purpose of `node_confirms` is to offer a guarantee that the data is available on multiple machines, for example setting node_confirms to 2 will guarantee that at least two machines have the data - and the risk of the data being lost can be considered accordingly.
+
+the `node_confirms` property were added as an alternative to using the `pr` and `pw` parameters to provide guarantees that distinct nodes were used for storage before confirming reads/writes.  The use of `pw` could both fail to provide the desired guarantee of physical redundancy; but also could prompt false failures in circumstances where the data was still resiliently stored, reducing availability.
```
ThomasArts:
```suggestion
The `node_confirms` property was added as an alternative to using the `pr` and `pw` parameters to provide guarantees that distinct nodes were used for storage before confirming reads/writes.  The use of `pw` could both fail to provide the desired guarantee of physical redundancy; but also could prompt false failures in circumstances where the data was still resiliently stored, reducing availability.
```

docs/BucketProperties.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `pr` and `pw` bucket properties default to `0`, and are used to require primary vnodes to be involved in reads and writes.  This may prevent writing to minority partitions, however when `{n_val, 3}` this will probably lead to intermittent failures when only two nodes fail in a cluster.  As clusters grow the probability of two concurrent failures will increase significantly.
+
+It is strongly recommended to consider using `node_confirms`, `sync_on_write` or token-based conditional PUTs to achieve controls in preference to configuring `pr`/`pw` to values greater than 1.
+
+It is normally best practice to configure either `{pr, 1}` or `{notfound_ok, false}`, rather than rely on defaults.  Other wise there is a potential issue when at least two nodes have failed and for some objects 2 of the 3 vnodes are unpopulated fallbacks.  In this case, without changing defaults, the two unpopulated fallback vnodes can return `not_found` and the GET request can achieve quorum and return a false not_found to the client.  By configuring either `{pr, 1}` or `{notfound_ok, false}`, when there is only one populated/primary vnode, the GET request must wait for this vnode to respond.
```
ThomasArts:
```suggestion
It is normally best practice to configure either `{pr, 1}` or `{notfound_ok, false}`, rather than rely on defaults.  Otherwise there is a potential issue when at least two nodes have failed and for some objects 2 of the 3 vnodes are unpopulated fallbacks.  In this case, without changing defaults, the two unpopulated fallback vnodes can return `not_found` and the GET request can achieve quorum and return a false not_found to the client.  By configuring either `{pr, 1}` or `{notfound_ok, false}`, when there is only one populated/primary vnode, the GET request must wait for this vnode to respond.
```

ThomasArts (approved):

ThomasArts (approved):

PR #51, opened 2025-09-03
Nhse o34 leveled.i433 enhancequery

martinsumner:
New Query API for Riak, intended to replace the existing use of secondary indexes.

The API will be usable via HTTP only in this release, and only supported in the Erlang HTTP client - https://github.com/OpenRiak/riak-erlang-http-client/pull/2.

Outstanding work is documentation related:
- Provide a much simpler example, and offer it first;
- Add performance information based on test rig and population sample data;
- Add examples for PUT (i.e. simple examples of how to update indexes).

[Riak Test](https://github.com/OpenRiak/riak_test/pull/12).


docs/QueryAPI.md
```diff
@@ -0,0 +1,592 @@
...
+```json
+    {
+        "substitutions" : {"dl1" : "|", "dl2" : ".", "qfn" : "SMITH", "qgn" : "ANNE"},
+        "query_list" :
+            [
+                {
+                    "index_name" : "peoplefinder_bin",
+                    "start_term" : "19650501",
+                    "end_term"   : "19650501~",
+                    "evaluation_expression" : "delim($term, :dl1, ($dob, $fn, $gn, $pc)) | split($gn, :dl2, $gn)",
```
ThomasArts, line 109: `split($gn, :dl2, $gn)` comes out of nowhere here.
Is worth explaining in the text below. I can figure out that you pick up the given name in `$gn`,  and then it looks like you split the given names with a dot? But from your example it is not clear that given names are presented as ANNA.MARIE.ANNA-MARIE in the index. And if that's the case, then I wonder what split actually does, return a list of those 3 names in the same variable `$gn`?

Right, split is explained below, I was wondering before reading. So what misses is whether these names are "." separated in the index

docs/QueryAPI.md
```diff
@@ -0,0 +1,592 @@
...
+    }
+```
+
+The evaluation expression is extended to output the birthday by taking the last 4 characters of the date of birth.  The filter expression checks an inexact match by looking only at the start of the family name.
+
+Alternative approaches would be possible:
+
+- `ends_with($dob, :qbd)` could be used for the birthday check avoiding the additional pipeline function in the evaluation expression.
+- `index($fn, 0, 2, $fn)` could be used in the evaluation expression to slim the $fn to the first two characters for equality checking.
+
```
ThomasArts: Would it also make sense to limit the date range, i.e., start_term "19650501" and end_term "19690502". Surely not many values we don't need to consider, but it is also an approach.

docs/QueryAPI.md
```diff
@@ -0,0 +1,592 @@
...
+### Consistency
+
+Index changes are not deferred to an async process, at a vnode level all index changes are made as a transaction with the object change.  Outside of failure scenarios, secondary index queries will almost always immediately reflect the results of any changes in the object (with caveats related to unreliable latency across intra-cluster networking communication).
+
+In failure and recovery scenarios, false negatives are possible (i.e. results may be missing until anti-entropy mechanisms correct) but results will be eventually consistent.  The query uses a coverage plan which will check only one (of N) potential copies of the data, and so should a vnode be temporarily incorrect, the entropy is not detected as part of the query.  The `participate_in_coverage` configuration option (which can be applied at run-time) us used to mitigate this - this can be used to prevent a node with a known entropy issue from being involved in queries.
```
ThomasArts:
```suggestion
In failure and recovery scenarios, false negatives are possible (i.e. results may be missing until anti-entropy mechanisms correct) but results will be eventually consistent.  The query uses a coverage plan which will check only one (of N) potential copies of the data, and so should a vnode be temporarily incorrect, the entropy is not detected as part of the query.  The `participate_in_coverage` configuration option (which can be applied at run-time) is used to mitigate this - this can be used to prevent a node with a known entropy issue from being involved in queries. 
```

src/riak_client.erl
```diff
@@ -1028,14 +1031,32 @@ hotbackup(BackupPath, DefaultNVal, PlanNVal, {?MODULE, [Node, _ClientId]}) ->
...
+-type query_result()
+    ::
+        [riak_object:key()] |
+        [{binary(), riak_object:key()}] |
+        #{binary() => non_neg_integer()} |
+        binary().
+
+-spec query(
+    riak_kv_query:complex_query_definition(), riak_client()) ->
+        {query_result(), none|{{binary(), riak_object:key()}}} |
```
ThomasArts, line 1049: `{{binary(), riak_object:key()}}` is a bit of a weird type. As if you make a singelton too many of the result.
Since eqwalizer does not complain, it might be the right type, but it may also have sneaked in.

src/riak_kv_query.erl
```diff
@@ -0,0 +1,1001 @@
...
+    ]
+).
+
+-spec new(riak_object:bucket(), query_type(), pos_integer()) -> complex_query_definition().
+new(Bucket, single_query, Timeout) ->
+    #riak_kv_query{
+        bucket = Bucket,
+        type = single_query,
+        timeout_secs = Timeout,
+        aggregation_expression = single_query
```
ThomasArts, line 186: since the default of `#riak_kv_query.aggregation_expression`  is `single_query` one could leave out this explicit setting and then the next `combo_query` and this one could be collapsed to `new(Bucket, QType, Timeout) ->` and set `type` to QType.
Not doing so, suggests that aggregation expression is something special for single_query, but then having it as a default for `combo_query` feels a bit weird.

src/riak_kv_query.erl
```diff
@@ -0,0 +1,1001 @@
...
+    complex_query_definition()) -> query_definition().
+get_query_definition(#riak_kv_query{type = Type} = Query)
+        when Type == single_query ->
+    case Query#riak_kv_query.query of
+        EvaluatedQuery when is_tuple(EvaluatedQuery) ->
+            EvaluatedQuery
+    end;
+get_query_definition(
+    #riak_kv_query{type = Type, aggregation_expression = AE} = Query)
+        when Type == combo_query, AE =/= single_query, AE =/= none ->
```
ThomasArts: Type tools did not pickup on this, but `none` is never a value for AE. aggergated_expression is aither single_query or a function.
It probably should be a value, that would make the above remark on combo_type easier, viz, add `none` there for aggregated_expression. So here's an inconsistency.

src/riak_kv_query.erl
```diff
@@ -0,0 +1,1001 @@
...
+add_maxresults(#riak_kv_query{accumulation_option = AccOpt, type = T} = Q, MR)
+        when is_integer(MR), MR > 0 ->
+    case {AccOpt, T} of
+        {AccOpt, single_query} when AccOpt == raw_keys; AccOpt == terms ->
+            {ok, Q#riak_kv_query{max_results = MR}};
+        {MaybeBadAccOpt, MaybeBadType} ->
+            {
+                error,
+                max_results,
+                list_to_binary(
```
ThomasArts: should this not be `iolist_to_binary` ?

src/riak_kv_query.erl
```diff
@@ -0,0 +1,1001 @@
...
+-spec make_continuation(index_limiter(), riak_object:key()) -> string().
+make_continuation(StartTerm, StartKeyExclusive) ->
+    M = #{?CONT_ST => StartTerm, ?CONT_SK => StartKeyExclusive},
+    base64:encode_to_string(iolist_to_binary(riak_kv_wm_json:encode(M))).
+
+-spec validate_substitutions(substitutions()) -> ok|validation_error().
+validate_substitutions(Subs) ->
+    try
+        maps:foreach(
+            fun(K, V) when is_binary(K), is_binary(V); is_integer(V) -> ok end,
```
ThomasArts: Not sure this is really what you want to express:
```
5> F = fun(K, V) when is_binary(K), is_binary(V); is_integer(V) -> ok end.
#Fun<erl_eval.41.105768164>
6> F(12,12).
ok
```

ThomasArts: Possibly you want 

```suggestion
            fun(K, V) when is_binary(K), is_binary(V) orelse is_integer(V) -> ok end,
```

rebar.config
```diff
@@ -1,4 +1,4 @@
-{minimum_otp_vsn, "22.0"}.
+{minimum_otp_vsn, "24.0"}.
```
ThomasArts, line 1: since you want to merge into openriak-3.4, would OTP26 not be the minimum?

src/riak_kv_query_sup.erl
```diff
@@ -0,0 +1,59 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_query_fsm_sup: supervise the riak_kv query state machines.
```
ThomasArts: This name is wrong
```suggestion
%% riak_kv_query_sup: supervise the riak_kv query state machines.
```

src/riak_kv_query_buffer.erl
```diff
@@ -0,0 +1,599 @@
...
+        rawkey_acc = none, key_acc = none, term_acc = none, rawterm_acc = none,
+        count = C
+    } = B) ->
+    B#buffer{count = C + 1};
+add(Key, #buffer{rawkey_acc = A, count = C} = Buffer)
+        when A =/= none, is_binary(Key) ->
+    Buffer#buffer{rawkey_acc = [Key|A], count = C + 1};
+add(Key, #buffer{key_acc = A, count = C} = Buffer)
+        when A =/= none, is_binary(Key) ->
+    Buffer#buffer{key_acc = [{Key}|A], count = C + 1};
```
ThomasArts, line 165: Why these `{ .. }` in the key_acc and later in term_acc. Would it not make sense to do the same as for raw key and raw term and then use the fact that you know what the type is later on?
I'm probably missing something here.

src/riak_kv_query_buffer.erl
```diff
@@ -0,0 +1,599 @@
...
+        term_acc = Acc, agg = #termcount_agg{map = TCMap}, type = T} = Buffer
+    )
+        when
+            Acc =/= none,
+            is_map(TCMap),
+            T == term_with_rawcount ->
+    reply(Buffer, ping),
+    UpdTCMap =
+        lists:foldl(
+            fun({{Term, _Key}}, FoldAcc) when is_map(FoldAcc), is_binary(Term) ->
```
ThomasArts, line 224: Here is really does not matter whether there is a `{{}}` or just one level of `{}`?

src/riak_kv_query_buffer.erl
```diff
@@ -0,0 +1,599 @@
...
+        end,
+    reply(Buffer, {T, Result}).
+
+-spec aggregate(reply_type(), reply_type()|none) -> reply_type().
+aggregate(R, none) ->
+    R;
+aggregate({T, KL}, {T, AggKL}) when T == keys ->
+    {T, lists:umerge(lists:usort(KL), AggKL)};
+aggregate({T, TKL}, {T, AggTKL}) when T == terms ->
+    {T, lists:umerge(lists:usort(lists:reverse(TKL)), AggTKL)};
```
ThomasArts: reverse here makes no sense, right? You sort it afterward and sort is not on a key.

```suggestion
    {T, lists:umerge(lists:usort(TKL), AggTKL)};
```

src/riak_kv_query_server.erl
```diff
@@ -0,0 +1,678 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_query_worker: Manage complex secondary index query.
```
ThomasArts: riak_kv_query_server? or is there a different convention than module name?

src/riak_kv_query_server.erl
```diff
@@ -0,0 +1,678 @@
...
+        code_change/3,
+        format_status/1
+    ]
+).
+
+-include_lib("kernel/include/logger.hrl").
+
+-define(VERSION, [{version, 2}]). % to be used in sets
+-define(ETS_THRESHOLD, 4096).
+% -define(START_OPTS, [{spawn_opt, [{min_heap_size, ?MIN_HEAP_SIZE}]}]).
```
ThomasArts: Is this a left over from tests, or for experiments. Add a comment why this is an alternative or leave it out completely.

src/riak_kv_wm_index.erl
```diff
@@ -584,11 +584,13 @@ otp_encode_results(false, Results, Continuation) ->
     ).
 
 results_encode({Term, Key}, Encode) when is_binary(Term), is_binary(Key) ->
-    ["{", [Encode(Term, Encode), $: | Encode(Key, Encode)], "}"];
+    [123, [Encode(Term, Encode), $: | Encode(Key, Encode)], 125];
```
ThomasArts: would you not rather use `${`, since we don't all have the ascii table in our head? And you already do it for `:`.
```suggestion
    [${, [Encode(Term, Encode), $: | Encode(Key, Encode)], $}];
```

src/riak_kv_wm_index.erl
```diff
@@ -584,11 +584,13 @@ otp_encode_results(false, Results, Continuation) ->
...
 results_encode({Term, Key}, Encode) when is_integer(Term), is_binary(Key) ->
-    ["{",
+    [
+        123,
```
ThomasArts: same here

src/riak_kv_query_server.erl
```diff
@@ -0,0 +1,678 @@
...
+                    max_results = MaxResults,
+                    result_encoding_fun
+                        = riak_kv_query:get_result_encodingfun(Query)
+                }
+            }
+    end.
+
+handle_cast({new_timeout, SecondsToTimeout}, State) ->
+    ReqID = riak_kv_query:get_reqid(),
+    erlang:send_after(SecondsToTimeout * 1000, self(), {timeout, ReqID}),
```
ThomasArts: This seems tricky.
You compute a new ReqId for the new timeout, but you don't clear the ongoing timeout. That's a nice way to avoid a race.

However, in the case your old timer goes of, I think there still is a message coming in `{timeout, OldReqId}`. But that will not match on line 253 below. Hence it falls all the way through and the catch all case stops the server, which is likely not what you intended.

I suggest a second handle info to ingnore timeouts that have the wrong ReqId

ThomasArts: Would be good to also have a test for this to show that extending the timeout by setting a new one works.

src/riak_kv_clusteraae_fsm.erl
```diff
@@ -78,7 +85,7 @@
     %% since unix epoch
 -type hash_method() :: pre_hash|{rehash, non_neg_integer()}.
     %% clocks are pre-hashed before storage to reduce CPU load for hash
-    %% comparisons.  However, there maye be hash collisions, and in this case
+    %% comparisons.  However, there may be be hash collisions, and in this case
```
ThomasArts:
```suggestion
    %% comparisons.  However, there may be hash collisions, and in this case
```

ThomasArts (approved):

src/riak_client.erl
martinsumner: This is the correct type.  The positive reply is {query_result(), LastResult}.  If max_results is not set, then LastResult will be `none`, however is max_results is set the LastResult needs to be sent, so that can be used as the starting point for the next query (i.e. continuing a paginated query).

The LastResult, needs to be both the IndexTerm (a binary()) and the object_key(), as both are required to define the starting point when requesting the next page.

docs/QueryAPI.md
martinsumner: Updated as suggested

docs/QueryAPI.md
martinsumner: Extra bullet added

docs/QueryAPI.md
martinsumner: Updated to add more details.

There is a lot to take in fast for the reader though.  The queries could build in complexity more gradually, but hard to get the balance as we want the docs to be relatively brief.  This is at 5K words, and there's still a significant TODO section.

May return to this as part of the 3.4 go-live document review.

src/riak_kv_query.erl
martinsumner: Removed the `none`

src/riak_kv_query.erl
martinsumner: I don't know, the only difference I can see is that `iolist_to_binary/1` expects `iolist()` and `binary()` where for `list_to_binary/1` it is just `iolist()`.  Though confusingly `io_lib:format/2` will output `chars()` not `iolist()`.

I don't think `chars` can be binary(), certainly not in this case - so `list_to_binary/1` is ok?

Does it matter either way?

martinsumner: Will use `iolist_to_binary/1` as it is more permissive

src/riak_kv_query_server.erl
martinsumner: Agree, this is broken.

I think it is there as I had an intention to offer some capability to poll, so that a query wouldn't timeout as long as requesting service was still waiting.  That wasn't done though.

I will remove the `new_timeout/2` API call and related stuff.  Will still check that the timeout ReqID is what was expected.

rebar.config
martinsumner: No,  Riak 3.4 release to initially support both OTP 24 and OTP 26 - so users can transition without a change of erlang version

src/riak_kv_query_buffer.erl
martinsumner: I think the list is generally in reverse order.  So perhaps I assumed it was quicker than reverse and sort (guessing that sorting a sorted list was quicker than sorting an unsorted list).

Done some experiments, and that doesn't seem to be true.  For every big lists `lists:reverse/1` is fairly expensive.

In fact sorting a reverse-sorted list is quicker than sorting a sorted list!  but not worth the effort of reversing it first.

Your suggestion looks good

src/riak_kv_query_buffer.erl
martinsumner: This is related to the use of ets for sorting.  Treating them this way means that they cna be handled the same when adding them into ets.

https://github.com/OpenRiak/riak_kv/commit/85f32c5d2c42f769ddfbdb705087683737d0f0cd

For raw key and raw term, they're not going to be inserted into ets for sorting, and so this doesn't matter.

I can't remember the details, but I do recall it was very deliberate.

src/riak_kv_query_buffer.erl
martinsumner: See previous comment

src/riak_kv_query.erl
martinsumner: Changes, and test added

src/riak_kv_query_buffer.erl
martinsumner: Looking back at other commit references, the issue is that the tuple wrapping is required for ets, but also riak_kv_query_server becomes the bottleneck when its sorting.  This is why the tuple wrapping happens at the buffer - so that an additional lists:map is not required to be done on the query_server

Issue #47, opened 2025-08-15
Two vnodes in the same preflist with the same vnode_id - amnesia chaos


martinsumner:
Copied from https://github.com/basho/riak_kv/issues/1865

The assigning of vnode ID is based on node ID and timestamp:

https://github.com/basho/riak_kv/blob/develop-3.0/src/riak_kv_vnode_status_mgr.erl#L342-L354

Vnodes on the same node are started in parallel, and so can have the same node ID and timestamp.  If the cluster is too small for claim to be effective, then two vnodes within the same preflist may be started with the same vnode ID.  In this circumstance, and write of a key co-ordinated by either vnode will cause key-amnesia then read repair will be prompted by that key amnesia in a continuous and never-ending loop.

There is potentially no recovery from this situation.

martinsumner:
I've personally only seen this once when running riak_test.  However, as computers get faster it may occur more frequently in the future.

martinsumner:
The `assign_vnodeid/3` function is puzzling:

https://github.com/OpenRiak/riak_kv/blob/644d04efd9e739504312dc8f79c1b753ea2595dd/src/riak_kv_vnode_status_mgr.erl#L344-L356

the NowEpoch is likely to be much bigger than `2 ^ 32`, as Secs is `0..1000000` and `4295 * 1000000 > 2 ^ 32`.  So the 32-bit integer added to the VnodeID is actually the least significant 32-bits of the integer.  So the VnodeID may not be different between restart, or indeed different to other IDs.

Why did it do this?  Looking back at much older code (prior to the introduction of riak_kv_vnode_status_mgr - as that function existed prior to that change) there is a comment on the call to `assign_vnodeid/3`:

```erlang
%% using now as we want different values per
%% vnode id on a node
assign_vnodeid(erlang:now(), riak_core_nodeid:get(), Status);
```

So the implication is that now is a just a bad way to get randomness, and the `VnodeEpoch = erlang:max(NowEpoch, LastVnodeEpoch+1)` is just confusion, rather than an attempt at making something monotonically increasing?

martinsumner:
The NodeID is a 4-byte binary consisting of a CRC check to the `term_to_binary/1` of the erlang node id.  This makes the VnodeID 8 bytes in length.

The preferred solution is to make the NodeID

```erlang
{Mega, Sec, _Micro} = Now,
NowEpoch = 1000000 * Mega + Sec,
[...]

PR #56, opened 2025-09-12
Use atomic module for unique vnode IDs

martinsumner:
See https://github.com/OpenRiak/riak_kv/issues/47


WarpEngineer (approved):

tburghart:
Does this really need a separate module?
If you trim down my commentary and build the functionality into where it's used, it'd probably only be a dozen lines of code and a couple of lines of comments.
I'll see if I can bang out a revision built off your modified `riak_kv_vnode_status_mgr` that keeps it to just that module.

tburghart:
Single-module alternate in PR #59

Issue #49, opened 2025-08-15
Handoff reaps - handling reap queues when there is cluster admin activity or failure


martinsumner:
Related to https://github.com/basho/riak_kv/issues/1850 & https://github.com/basho/riak_kv/pull/1851

tombstones exist, primarily when we use `delete_mode = keep`.  The use of `delete_mode = keep` is now the recommended way of handling deletes when running multi-cluster environments.

Garbage collection of tombstones is eventually required (as in large numbers they will impact performance regardless of backend choice - at a backend they still represent a key).  The `riak_kv_reaper` and the `reap_tombs` `aae_fold` were added to Riak KV to help with this garbage collection.  This was then extended to allow for reap requests to be replicated in nextgenrepl replication - so that clusters can be kept in sync during reaping.

In healthy clusters, an operator can now prompt a release job.  The reaps will be replicated through the cluster and linked clusters, and the clusters will stay in sync.  However, this isn't true in three scenarios:

- Where a sink cluster (receiving replicated reap requests) cannot apply the reaps as fast as the source cluster;
- Where a sink cluster (receiving replicated reap requests) is undergoing a cluster admin operation (e.g. a join, leave, replace or repair);
- Where a sink cluster (receiving replicated reap requests) has a node down.

In all these scenarios the clusters will become out of sync, nexgenrepl full-sync (and in the latter two cases internal AAE) will spot discrepancies, and start to work to resurrect tombstones.  There can be situations where a large reap job, combined with a cluster change, may lead to o(days) full-sync activity.

There are some workarounds at present.  

The `tombstone_pause` is configurable: https://github.com/OpenRiak/riak_kv/blob/644d04efd9e739504312dc8f79c1b753ea2595dd/priv/riak_kv.schema#L342-L352

This pause can be configured in linked clusters so that "slow" clusters have a low pause, and "fast" clusters have a high pause - so that "slow" clusters don't fall behind.  For extra safety operators should also prompt reap jobs on slow clusters.

Manipulating the `tombstone_pause` does not help in the admin operation of node down scenario.  In these cases, the only workaround is for operators to pause such reap jobs when a failure occur , and suspend them during cluster change windows.

In the future it would be preferred to allow for sweeper-like garbage collection of tombstones to ease the operator burden - so relying on operator intervention would not be possible.  This is a hard problem, as the issue occurs when the failure/change is an a cluster other than the one that is controlling the reap work

martinsumner:
[...]

PR #57, opened 2025-09-17
Reap on all available primaries

martinsumner:
Defer the reap only for unavailable primaries.

Previously the reap was deferred on all primaries if some were not available.  This caused issues with multi-cluster environments - where a set of reaps would occur on one cluster, but not on another (due to a node being unavailable).  This caused full-sync reconciliation issues and the resurrection of tombstones.

### DIFF
--- a/src/riak_kv_reaper.erl
+++ b/src/riak_kv_reaper.erl
@@ -183,7 +183,7 @@ action({{Bucket, Key}, DeleteHash, Indices}, Redo)
             %% But pause first - so that the process is not in a tight loop
             %% re-checking
             timer:sleep(TombPause),
-            maybe_redo(Redo);
+            respond_asfail_ifredo(Redo);
         {Available, Deferred} when Available =/= [] ->
             case check_all_mailboxes(Available) of
                 ok ->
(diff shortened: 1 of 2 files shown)


martinsumner:
This covers the failure aspect of https://github.com/OpenRiak/riak_kv/issues/49.

Handling reaps during admin changes is more complex, and so will be deferred.

src/riak_kv_reaper.erl
```diff
@@ -190,6 +262,64 @@ redo() -> true.
...
+            N,
+            riak_core_node_watcher:nodes(riak_kv)
+        )
+    }.
+
+%% @doc
+%% Redo by indicating the original request was not successful (false)
+-spec maybe_redo(boolean()) -> boolean().
+maybe_redo(Redo) ->
+    if Redo -> false; true -> true end.
```
ThomasArts: This is a boolean negation. Given the type `boolean() -> boolean()` you could as well write as body `not Redo`.
Which might be easier to read for some people.

src/riak_kv_reaper.erl
```diff
@@ -150,37 +164,95 @@ get_limits() ->
...
+                    timer:sleep(TombPause),
+                    ok = 
+                        riak_kv_queue_manager:redo_request(
+                            ?MODULE,
+                            {{Bucket, Key}, DeleteHash, Deferred}
+                        ),
+                    %% Some indices were available and the cluster is not
+                    %% stressed - so some updates made, and the remaining are
+                    %% deferred
+                    true;
```
ThomasArts, line 208: since you use `maybe_redo()` for other returns, would it make sense to use `redo()` here?

ThomasArts (approved):

src/riak_kv_reaper.erl
martinsumner: Updated the code, and the comments to make it clearer why negation is happening

src/riak_kv_reaper.erl
martinsumner: The redo function has a specific role as a callback in the behaviour.  Clarified comments and function names, so hopefully this is now clearer

PR #61, opened 2025-09-30
Nhse o34 orkv.i38 docsanddefaults

martinsumner:
Doc update and default changes

### DIFF
--- a/docs/InitialDesignDecisions.md
+++ b/docs/InitialDesignDecisions.md
@@ -1,0 +1,122 @@
+# Initial Design Decisions when Using Riak
+
+When starting with Riak a number of initial design decisions need to be made at the outset of the project.  This is a summary of those decisions, and the factors relevant to making each choice.
+
+The initial design decisions are split into the following categories:
+- Database backend
+- Ring size
+- Anti-entropy and reconciliation
+- Replication and data resilience
+- Bucket properties
+- Mapping data to objects
+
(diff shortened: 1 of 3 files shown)


martinsumner:
Other changes in https://github.com/orgs/OpenRiak/discussions/26 - are pending further discussion

WarpEngineer:
Hey @martinsumner I'm not convinced we should make these breaking changes going from 3.2 to 3.4.  It's not a big enough version change for that.  I think we should save this for version 4.0.

martinsumner:
The ones I've added to this PR so far, I thought, were the relatively non-controversial changes (compared to changing the `allow_mult` or `notfound_ok` defaults).

The `handoff_deletes` is a recent addition, which was required for one Riak user, but kept as default of off when it was introduced until it has been through production proving with that user.

Also there's the question of timescales, as to when a major (e.g. 4.0) version will actually be available.

Perhaps one to discuss at the next catch-up, I'm open to be persuaded.

WarpEngineer:
Yeah, let's discuss again at the catch-up.

martinsumner:
@WarpEngineer - I've reverted out the `dvv_enabled` change (and updated the linked issue to indicate this will be done in 4.0).  Ar you OK to go forward with just the change to handoff_deletes and the doc update?

WarpEngineer (approved):

PR #65, opened 2025-10-22
Convert riak object metadata into a map

martinsumner:
The map module is more performant that the dict module, in part as dict is coupled to an old (deprecated) hash algorithm.

All objects are stored in binary form - which is independent of the data type.  This code will still handle objects converting metadata to dict.  However, enabling the switch to version 1 will require all connected nodes to support metadata as map.

A capability could be used, but there is history of capability registrations taking a long time to remove - so the shortcut of making this an operator responsibility has been made, especially with a major version change pending.


martinsumner:
Tested - https://github.com/OpenRiak/riak_test/pull/14

Also requires - https://github.com/OpenRiak/riak-erlang-http-client/pull/3 and https://github.com/OpenRiak/riak_pb/pull/2

Also solves - https://github.com/OpenRiak/riak_kv/issues/70

WarpEngineer (approved):

WarpEngineer (approved):

Issue #68, opened 2025-10-31
`delete_mode` configuration - add to `riak.conf`


martinsumner:
The `delete_mode` is an important configuration option, yet it is not configurable within `riak.conf`.

The reasons why it wasn't added previously aren't clear, perhaps there were issues due to the different types supported in the resultant value?

martinsumner:
Making the change doesn't work when building Riak:

```
Error: badarg
[{io_lib,format,
         ["Default: ~s",[{error,{type,{3000,atom}}}]],
         [{file,"io_lib.erl"},{line,189}]},
 {cuttlefish_conf,generate_comments,1,
                  [{file,"/Users/martinsumner/dbroot/OpenRiak/riak/_build/default/plugins/cuttlefish/src/cuttlefish_conf.erl"},
                   {line,217}]},
 {cuttlefish_conf,generate_element,1,
                  [{file,"/Users/martinsumner/dbroot/OpenRiak/riak/_build/default/plugins/cuttlefish/src/cuttlefish_conf.erl"},
                   {line,185}]},
 {cuttlefish_conf,'-generate/1-fun-0-',2,
                  [{file,"/Users/martinsumner/dbroot/OpenRiak/riak/_build/default/plugins/cuttlefish/src/cuttlefish_conf.erl"},
                   {line,135}]},
 {lists,foldl_1,3,[{file,"lists.erl"},{line,1599}]},
 {cuttlefish_conf,generate_file,2,
                  [{file,"/Users/martinsumner/dbroot/OpenRiak/riak/_build/default/plugins/cuttlefish/src/cuttlefish_conf.erl"},
                   {line,140}]},
 {rebar3_cuttlefish_release,make_default_file,3,
                            [{file,"/Users/martinsumner/dbroot/OpenRiak/riak/_build/default/plugins/rebar3_cuttlefish/src/rebar3_cuttlefish_release.erl"},
                             {line,130}]},
 {rebar3_cuttlefish_release,do,1,
                            [{file,"/Users/martinsumner/dbroot/OpenRiak/riak/_build/default/plugins/rebar3_cuttlefish/src/rebar3_cuttlefish_release.erl"},
                             {line,113}]}]
```

Issue appears to be that if a list of potential types is provided, the first in the list must be the type used by the default

PR #72, opened 2025-11-07
Add delete_mode to schema, with some tests of "invalid" values

martinsumner:
In theory any positive integer is possible for timeout.  But deleting 1000s a second would require a lot of timer process to be spawned if the limit was very high?

Limiting to some (hopefully) sane limit - but allowing it to be overridden in advanced.config (as with compromise on configuration of large ring_size).


martinsumner:
https://github.com/OpenRiak/riak_kv/issues/68

WarpEngineer (approved):

PR #75, opened 2025-11-13
Nhse o34 orkv.i73 quickdocs2

martinsumner:
Review of spelling and grammar, assisted by google docs


docs/BuildAndScaleClusterGuide.md
```diff
@@ -104,7 +119,7 @@ With modern hardware, a simple configuration such as this can achieve a very hig
...
+A cluster is formed by joining nodes to a cluster.  Note that a Riak node when it is started, is a cluster of one, and so the acting of joining a node is actually the act of merging two clusters.  If the ring size is 256, a Riak node that is not part of a cluster will start 256 vnodes as it considers itself to be the whole cluster.  When nodes join a cluster, the handoff process is two-ways - the joining node is handing off vnodes it will no longer run to the cluster, and the cluster will hand off vnodes it requires the joining node to run to that node.  Note that each vnode consists of two vnode modules - `riak_kv_vnode` and `riak_pipe_vnode` - and both modules must handoff for a vnode handoff to complete (although generally the `riak_pipe_vnode` is empty so this handoff is immediate).
```
WarpEngineer: 'The act of joining a node' instead of 'the acting of joining a node'.

docs/InitialDesignDecisions.md
```diff
@@ -57,12 +57,13 @@ The leveled backend has the following characteristics and features:
 
 The bitcask backend has the following characteristics and features:
 
-- A simple low-code, journal based key-value store, originally design and developed by the Riak team; that has been used as a general implementation model for other such stores.
+- A simple low-code, journal based key-value store, originally designed and developed by the founding Riak team at Basho; that has been used as a general implementation model for other such stores.
+  - Although there have been minimal changes to the original version as at Riak 3.4, bitcask remains under active development by the OpenRiak team, with plans agreed for future enhancements.
```
WarpEngineer:
```suggestion
  - Although there have been minimal changes to the original version, as of Riak 3.4, bitcask remains under active development by the OpenRiak team, with plans agreed for future enhancements.
```

docs/InitialDesignDecisions.md
```diff
@@ -179,29 +180,31 @@ It is possible to control the possibility of consistency issues using conditiona
...
+The complexity of deletion with eventual consistency is that it is an underlying requirement in a distributed system to compare potentially differing results between different locations for an object (i.e. between vnodes or between clusters).  If location A has an object, and location B doesn't; there is a need to differ between the situation where location B is correct (due to a deletion not being replicated), or location A is correct (due to an insertion not being replicated) - as these circumstances require opposing actions.  There are secondary consequences if keys are reused following deletion, whereby data could be potentially lost if an old deleted object is resurrected and appears to be more recent.
```
WarpEngineer, line 203: This is fine but a bit hard to follow. Had to read it twice.

docs/InitialDesignDecisions.md
```diff
@@ -226,9 +229,9 @@ The delete mode is a per-node configuration which needs to be applied consistent
...
-- Eventually parallel writes will occur, and siblings will exist.  Siblings can be minimised using conditional PUTs, if sibling resolution is complex or requires manual intervention.  Use aae_folds feeding operator dashboards ot track the generation of siblings.  To auto-resolve siblings CRDTs (conflict-free replicated data types) can be used, and third-party client-side libraries are generally a better long-term option than using Riak's internal CRDTs.
+- Optimise the model for reading not writing, by storing information that is likely required to be fetched together in the same object.  It is normally easier to fetch a single object and strip unnecessary information, than it is to fetch multiple objects to fulfill a single data-demand from the application; where possible make the most common read requests fulfilled via a single object read request.
```
WarpEngineer:
```suggestion
- Optimise the model for reading not writing, by storing information that is likely required to be fetched together in the same object.  It is normally easier to fetch a single object and strip unnecessary information, than it is to fetch multiple objects to fulfill a single data-demand from the application. Where possible, make the most common read requests fulfilled via a single object read request.
```

docs/InstallAndStartGuide.md
```diff
@@ -33,7 +33,7 @@ Some points to note when installing Erlang:
 - The OpenSSL 3.0 integration in OTP 24 is not currently considered to be production-ready and stable.
 - If using Riak KV 3.0.16 and OTP 22.3, Riak does not support Erlang/OTP running in [HIPE mode](https://www.erlang.org/docs/22/man/hipe_app).  HIPE is retired as of OTP 24.
 - There are significant performance advantages in running Riak on OTP 26, when compared with OTP 24.3.
-- The Erlang/OTP team only commit to fix issues in the three most recent major versions of Erlang.  Although Erlang 24.3 is mature and very stable, migrating forward to a Riak release running on a presently supported Erlang version is recommended.
+- The Erlang/OTP are team only committed to fixing issues in the three most recent major versions of Erlang.  Although Erlang 24.3 is mature and very stable, migrating forward to a Riak release running on a presently supported Erlang version is recommended.
```
WarpEngineer:
```suggestion
- The Erlang/OTP team are only committed to fixing issues in the three most recent major versions of Erlang.  Although Erlang 24.3 is mature and very stable, migrating forward to a Riak release running on a presently supported Erlang version is recommended.
```

docs/InstallAndStartGuide.md
```diff
@@ -200,15 +200,15 @@ Unless the non-existence of an object can be guaranteed by the application using
...
 
-In general, the default should be used, even when `{allow_mult, false}`.  Setting `{last_write_wins, true}` changes the behaviour on PUT, so that an incoming write is assumed to be superior to an existing write without checking the change history of the existing object.  Internally within Riak, the actual order which PUTs are applied is non-deterministic, and there are many situations (replication, anti-entropy, handoffs) where old PUTs may be received after new PUTs.  In these cases setting `{last_write_wins, true}` may have unexpected consequences
+In general, the default should be used, even when `{allow_mult, false}`.  Setting `{last_write_wins, true}` changes the behaviour on PUT so that an incoming write is assumed to be superior to an existing write, without checking the change history of the existing object.  Internally within Riak, the actual order with which PUTs are applied is non-deterministic, and there are many situations (replication, anti-entropy, handoffs) where old PUTs may be received after new PUTs.  In these cases setting `{last_write_wins, true}` may have unexpected consequences
```
WarpEngineer:
```suggestion
In general, the default should be used, even when `{allow_mult, false}` is set.  Setting `{last_write_wins, true}` changes the behaviour on PUT so that an incoming write is assumed to be superior to an existing write, without checking the change history of the existing object.  Internally within Riak, the actual order with which PUTs are applied is non-deterministic, and there are many situations (replication, anti-entropy, handoffs) where old PUTs may be received after new PUTs.  In these cases setting `{last_write_wins, true}` may have unexpected consequences
```

docs/InstallAndStartGuide.md
```diff
@@ -224,15 +224,15 @@ Changing an n_val on a bucket which already contains data will have unexpected a
...
 
-the `node_confirms` property was added as an alternative to using the `pr` and `pw` parameters to provide guarantees that distinct nodes were used for storage before confirming reads/writes.  The use of `pw` could both fail to provide the desired guarantee of physical redundancy; but also could prompt false failures in circumstances where the data was still resiliently stored, reducing availability.
+The `node_confirms` property was added as an alternative to using the `pr` and `pw` parameters to provide guarantees that distinct nodes were used for storage before confirming reads/writes.  The use of `pw` could both fail to provide the desired guarantee of physical redundancy; but also could prompt false failures in circumstances where the data was still resiliently stored, reducing availability.
```
WarpEngineer:
```suggestion
The `node_confirms` property was added as an alternative to using the `pr` and `pw` parameters to provide guarantees that distinct nodes were used for storage before confirming reads/writes.  The use of `pw` could fail to provide the desired guarantee of physical redundancy and also could prompt false failures in circumstances where the data was still resiliently stored, reducing availability.
```

docs/NextGenReplGuide.md
```diff
@@ -246,31 +246,31 @@ In a cluster with 1bn keys, under a steady load including 2K PUTs per second, re
...
+The `ttaaefs_queuename` is the name of the queue on this node, to which deltas should be written (assuming the remote cluster being compared has sink workers fetching from this queue).  If the `ttaaefs_queuename_peer` is set to disabled, when repairs are discovered, but it is the peer node that has the superior value, then these repairs are ignored.  It is expected these repairs will be picked up instead by discovery initiated from the peer.  Setting the `ttaaefs_queuename_peer` to the name of a queue on the peer which this node has a sink worker enabled to fetch from will actually trigger repairs when the peer cluster is superior.  It is recommended to make full-sync repair bi-directionally in this way, as otherwise resources spent on reconciliation activity triggered by the cluster with inferior value will be wasted.
```
WarpEngineer:
```suggestion
The `ttaaefs_queuename` is the name of the queue on this node, to which deltas should be written (assuming the remote cluster being compared has sink workers fetching from this queue).  If the `ttaaefs_queuename_peer` is set to disabled when repairs are discovered, but it is the peer node that has the superior value, then these repairs are ignored.  It is expected these repairs will be picked up instead by discovery initiated from the peer.  Setting the `ttaaefs_queuename_peer` to the name of a queue on the peer which this node has a sink worker enabled to fetch from will actually trigger repairs when the peer cluster is superior.  It is recommended to make full-sync repair bi-directionally in this way, as otherwise resources spent on reconciliation activity triggered by the cluster with inferior value will be wasted.
```

docs/NextGenReplGuide.md
```diff
@@ -246,31 +246,31 @@ In a cluster with 1bn keys, under a steady load including 2K PUTs per second, re
...
 
-Note that a scheduled run of an `ttaaefs_allcheck` will occur regardless of whether the current time is within or outside of the allcheck.window.  The window is related only to the running of `ttaaefs_autocheck` and it limits the `ttaaefs_autocheck` so that it can only being uplifted to a `ttaaefs_allcheck` within the window, outside of the window it will only be ` ttaaefs_daycheck`.
+Note that a scheduled run of an `ttaaefs_allcheck` will occur regardless of whether the current time is within or outside of the `allcheck.window`.  The window is related only to the running of `ttaaefs_autocheck` and it limits the `ttaaefs_autocheck` so that it can only be uplifted to a `ttaaefs_allcheck` within the window.
```
WarpEngineer:
```suggestion
Note that a scheduled run of `ttaaefs_allcheck` will occur regardless of whether the current time is within or outside of the `allcheck.window`.  The window is related only to the running of `ttaaefs_autocheck` and it limits the `ttaaefs_autocheck` so that it can only be uplifted to a `ttaaefs_allcheck` within the window.
```

docs/NextGenReplGuide.md
```diff
@@ -246,31 +246,31 @@ In a cluster with 1bn keys, under a steady load including 2K PUTs per second, re
...
+There are significant memory improvements related to the new tree format, so the configuration should be reversed after the rolling upgrade has completed.  There are no inter-cluster issues with tree versions, it is only an issue when merging trees within a cluster to provide a cluster-wide view of a tree.
 
 ### Replication API
 
 All real-time and full-sync operations are available via the Riak API, and supported by the Riak erlang clients (both PB and HTTP).  They have been used in different systems for replicating and synchronising with third party databases, such as OpenSearch or DynamoDB.
 
-It is recommended, where possible to use the PB API for performance reasons, and also as TLS security can be enabled via this API.  The HTTP API is slower, but maybe useful where it is easier to set up peer relationships with a HTTP-based load-balancer rather than an individual node.
+It is recommended to use the PB API for both performance and security reasons (as TLS can be enabled via this API for replication).  The HTTP API is slower, but maybe useful where it is easier to set up peer relationships with a HTTP-based load-balancer rather than an individual node.
```
WarpEngineer:
```suggestion
It is recommended to use the PB API for both performance and security reasons (as TLS can be enabled via this API for replication).  The HTTP API is slower, but may be useful where it is easier to set up peer relationships with a HTTP-based load-balancer rather than an individual node.
```

docs/ObjectAPI.md
```diff
@@ -206,14 +206,14 @@ When using the HEAD method the request will still result in the object value bei
 Supported HTTP request headers for GET:
 
 - `authorization`; optional, for tls-protected requests only when riak security is enabled.
-- `accept: multipart/mixed`; optional, will cause results ina. conflicted state to return all siblings as one multipart-mime object body.  Without this option a list of sibling vtags will be returned, and each vtag mat be fetched using the `vtag=<vtag>` query parameter in the URL.
+- `accept: multipart/mixed`; optional, will cause results ina. conflicted state to return all siblings as one multipart-mime object body.  Without this option a list of sibling vtags will be returned, and each vtag may be fetched using the `vtag=<vtag>` query parameter in the URL.
```
WarpEngineer:
```suggestion
- `accept: multipart/mixed`; optional, will cause results in a conflicted state to return all siblings as one multipart-mime object body.  Without this option a list of sibling vtags will be returned, and each vtag may be fetched using the `vtag=<vtag>` query parameter in the URL.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -28,49 +28,53 @@ The most common repair requirements are for proactive replace, and reactive repl
...
+After completing a proactive replace operation, it may be necessary to realign node naming with design documents or monitoring systems; to rename a replacement node with the name of the node it replaced.  Once the replace operation is complete, it is possible to rename a node while it is down using `reip_manual` - see `riak admin reip_manual --help`.  The ring_directory required is normally named `ring` in the platform data directory.  It will contain files such as `riak_core_ring.default.20221122164111`, where the middle term between the period (in this case `default`) represents the required cluster name.
```
WarpEngineer:
```suggestion
After completing a proactive replace operation, it may be necessary to realign node naming with design documents or monitoring systems; to rename a replacement node with the name of the node it replaced.  Once the replace operation is complete, it is possible to rename a node while it is down using `reip_manual` - see `riak admin reip_manual --help`.  The ring_directory is normally named `ring` in the platform data directory.  It will contain files such as `riak_core_ring.default.20221122164111`, where the middle term between the periods (in this case `default`) represents the required cluster name.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -28,49 +28,53 @@ The most common repair requirements are for proactive replace, and reactive repl
...
 
 ### Reactive Replace
 
-When a node has failed following an incident, the node can still be recovered back to its previous state without requiring a backup.
+If a node temporarily fails, then recovers without a loss of historic delta; the node will automatically rejoin the cluster and have the any delta in data patched via ant-entropy mechanisms, without the need for operator intervention.
```
WarpEngineer:
```suggestion
If a node temporarily fails, then recovers without a loss of historic delta; the node will automatically rejoin the cluster and have any delta in data patched via ant-entropy mechanisms, without the need for operator intervention.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -89,49 +93,49 @@ riak eval "riak_client:remove_node_from_coverage()."
...
 
-To improve the performance of repair the `repair_span` configuration in the [riak_core schema section of riak.conf](https://github.com/OpenRiak/riak_core/blob/openriak-3.4/priv/riak_core.schema) can be changed to `double_pair`, and this has been proven to be more effective when used with the leveled backend and the enablement of the `repair_deferred` option in the [riak_kv schema section of riak.conf](https://github.com/OpenRiak/riak_kv/blob/openriak-3.4/priv/riak_kv.schema).
+To improve the performance of repair the `repair_span` configuration in the [riak_core schema section of riak.conf](https://github.com/OpenRiak/riak_core/blob/openriak-3.4/priv/riak_core.schema) can be changed to `double_pair`, and this has been proven to be more effective when used with the leveled backend together with the enablement of the `repair_deferred` option in the [riak_kv schema section of riak.conf](https://github.com/OpenRiak/riak_kv/blob/openriak-3.4/priv/riak_kv.schema).
```
WarpEngineer:
```suggestion
To improve the performance of repair, the `repair_span` configuration in the [riak_core schema section of riak.conf](https://github.com/OpenRiak/riak_core/blob/openriak-3.4/priv/riak_core.schema) can be changed to `double_pair`, and this has been proven to be more effective when used with the leveled backend together with the enablement of the `repair_deferred` option in the [riak_kv schema section of riak.conf](https://github.com/OpenRiak/riak_kv/blob/openriak-3.4/priv/riak_kv.schema).
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -89,49 +93,49 @@ riak eval "riak_client:remove_node_from_coverage()."
...
 
 There exists the (very rare) potential for a ledger to be corrupted.  There are also circumstances where following an update the ledger is not fully efficient until it is rebuilt.  As a missing ledger is rebuilt automatically on startup, rebuilding of the ledger can simply be prompted by deleting it:
 
 - Stopping the node;
 - Deleting the ledgers in the impacted partitions (under each vnode's leveled store there should be a ledger folder);
 - Restarting the node.
 
-On restarting the node all missing ledgers will be rebuilt before the node becomes and active participant in the cluster (the riak_kv application will not complete startup until the rebuilds are complete).  Rebuild progress can be tracked in the leveled logs with `log_ref=b0006`.
+On restarting the node all missing ledgers will be rebuilt before the node becomes and active participant in the cluster - the `riak_kv` application which determines availability of a node, will not complete startup until all the rebuilds are complete.  Rebuild progress can be tracked in the leveled logs with `log_ref=b0006`.
```
WarpEngineer:
```suggestion
On restarting the node all missing ledgers will be rebuilt before the node becomes an active participant in the cluster - the `riak_kv` application which determines availability of a node, will not complete startup until all the rebuilds are complete.  Rebuild progress can be tracked in the leveled logs with `log_ref=b0006`.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -264,26 +268,26 @@ Each worker pool will regularly log its current queue length and last checkout t
 
 ### Riak KV Reaper and Riak KV Eraser
```
WarpEngineer: Change this to mention the Eraser then the Reaper to be consistent with the order in the contents.

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -264,26 +268,26 @@ Each worker pool will regularly log its current queue length and last checkout t
...
+Likewise the `riak_kv_reaper` process receives requests to delete tombstones, queues those requests, and continuously reaps keys referenced in the queue.  Refer to the [API guide for AAE Fold](/docs/OtherAPI.md#aae-fold-api) for information on triggering a `reap_tombs` AAE fold to feed the reaper queue.
 
 Filters within the AAE folds can be used to select specific key_ranges, or last modified date ranges for the erase or reap process.
 
 When queueing large volumes of changes, note that:
 
-- The number of vnodes per node on which the fold is run will be restricted by the size of the `AF4_QUEUE` if the `dscp` worker strategy is used.  This will lead to a situation where items on the queue will be grouped by vnode, and the dequeued in batches by within the same preflist.
+- The number of vnodes per node on which the fold is run will be restricted by the size of the `AF4_QUEUE` if the `dscp` worker strategy is used.  This will lead to a situation where items on the queue will be grouped by vnode, and the dequeued in batches containing objects within the same preflist.
```
WarpEngineer:
```suggestion
- The number of vnodes per node on which the fold is run will be restricted by the size of the `AF4_QUEUE` if the `dscp` worker strategy is used.  This will lead to a situation where items on the queue will be grouped by vnode, and dequeued in batches containing objects within the same preflist.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -264,26 +268,26 @@ Each worker pool will regularly log its current queue length and last checkout t
...
 - In multi-data centre configurations reap events must be specifically configured to be replicated - this is controlled through the `repl_reap` configuration setting.  Otherwise reap jobs must be run separately on each cluster (with reconciliation suspended until the reaps complete).
 - Each queue on each node has a limit to the size of reaps or erases it can hold - this is controlled through the `eraser_overflow_limit` and the `reaper_overflow_limit`.  The queue, except for a small number, is held on disk; and so increasing this limit can be achieved without hitting memory constraints.
-- As of Riak 3.4, ff a reap is de-queued, but the primaries are not all available, then the reap will be acted on all available primaries and an item will be queued to act on the remaining primaries once they are available.
+- As of Riak 3.4, if a reap is de-queued, but the primaries are not all available, then the reap will be acted on all available primaries and an item will be queued to act on the remaining primaries once they are available.
```
WarpEngineer: I suggest 'dequeued' instead of 'de-queued' to be consistent with previous usage.

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -365,31 +369,35 @@ For performance testing Riak [`basho_bench`](https://github.com/OpenRiak/basho_b
...
 
 Note that in cloud environments, if an inefficient backup method is chosen (e.g. snapshots of block-service file-system volumes), then backup costs may consume a dominant proportion of overall Riak infrastructure costs.
 
 ### Backup - the preferred building block
 
-As part of the replication approach of Riak it is possible to replicate, and reconcile between clusters with different `n_val`'s, different node counts, different vnode_counts and different storage backends.  It is common in Riak production systems to maintain a single-node, `n_val=1`, potentially lower ring_size, cluster that represents a backup; where that cluster may be in a diverse geographical occasion to the primary production clusters.
+As part of the replication approach of Riak it is possible to replicate, and reconcile between clusters with different `n_val`s, different node counts, different vnode counts (i.e. ring sizes) and different storage backends.  It is common in Riak production systems to maintain a single-node, `n_val=1` cluster, with a potentially lower ring size, that represents a backup; where that cluster may be in a diverse geographical occasion to the primary production clusters.
```
WarpEngineer:
```suggestion
As part of the replication approach of Riak it is possible to replicate, and reconcile between clusters with different `n_val`s, different node counts, different vnode counts (i.e. ring sizes) and different storage backends.  It is common in Riak production systems to maintain a single-node, `n_val=1` cluster, with a potentially lower ring size, that represents a backup; where that cluster may be in a diverse geographical location to the primary production clusters.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -365,31 +369,35 @@ For performance testing Riak [`basho_bench`](https://github.com/OpenRiak/basho_b
...
+The real-time replication process that keeps the backup in-sync use a queue on the source cluster, and that queue will grow (as small on-disk references to changes) should the sink (i.e. the backup cluster) pause consumption.  After resuming replication, the sink cluster will catch-up on the missing changes from the queue, and when reconciliation is re-enabled that catch-up can be confirmed.  There is flexibility to disconnect a backup cluster, hold it at a point in time, and then in the future reconnect and fast-forward to the current state, then prove that the fast-forward was successful - with automatic resolution of any unexpected deltas.
```
WarpEngineer:
```suggestion
The real-time replication process that keeps the backup in-sync uses a queue on the source cluster, and that queue will grow (as small on-disk references to changes) should the sink (i.e. the backup cluster) pause consumption.  After resuming replication, the sink cluster will catch-up on the missing changes from the queue, and when reconciliation is re-enabled that catch-up can be confirmed.  There is flexibility to disconnect a backup cluster, hold it at a point in time, and then in the future reconnect and fast-forward to the current state, then prove that the fast-forward was successful - with automatic resolution of any unexpected deltas.
```

docs/QueryAPI.md
```diff
@@ -497,11 +501,11 @@ kvsplit ( IN_ID identifier , PAIR_DELIM string , KV_DELIM string )
...
 
-- to classify the value of an projected attribute the map function is used.  The MAP_LIST is a list of pairs, where the first element of the pair is a value to compare with, and the second element is a classification for a match against this  pair (the output value).  The comparison between the value and the first element is done using the comparator COMP.  If no element of the MAP_LIST returns a match against the input value, then the DEFAULT classification is used as the output value.  The output value is added to the projected attributes using the OUT_ID identifier as the key.
+- to classify the value of a projected attribute the map function is used.  The MAP_LIST is a list of pairs, where the first element of the pair is a value to compare with, and the second element is a classification for a match against this  pair (the output value).  The comparison between the value and the first element is done using the comparator COMP.  If no element of the MAP_LIST returns a match against the input value, then the DEFAULT classification is used as the output value.  The output value is added to the projected attributes using the OUT_ID identifier as the key.
```
WarpEngineer, line 508:
```suggestion
- to classify the value of a projected attribute, the map function is used.  The MAP_LIST is a list of pairs, where the first element of the pair is a value to compare with, and the second element is a classification for a match against this  pair (the output value).  The comparison between the value and the first element is done using the comparator COMP.  If no element of the MAP_LIST returns a match against the input value, then the DEFAULT classification is used as the output value.  The output value is added to the projected attributes using the OUT_ID identifier as the key.
```

WarpEngineer (changes requested):

docs/InitialDesignDecisions.md
martinsumner: at -> of works ... but I'm not sure about the addition of the comma

docs/NextGenReplGuide.md
martinsumner: Removing comma, but making disabled `disabled` to make it clear it is a setting

docs/OperationsAndTroubleshootingGuide.md
martinsumner: Also will fix "ant-entropy"

docs/QueryAPI.md
martinsumner: I prefer without the comma

WarpEngineer (approved):

Issue #77, opened 2025-11-18
Riak KV Query - info logs


martinsumner:
The Riak KV query generates a log of each query.  Should there be per-request logging?

This is an error, left in from troubleshooting.

PR #79, opened 2025-11-20
Nhse o34 or.i4 logger

martinsumner:
Switch to using domain [background, metric] for metric log


martinsumner:
Also covers https://github.com/OpenRiak/riak_kv/issues/77

WarpEngineer (approved):

PR #81, opened 2025-11-21
Nhse o34 orkv.i80 replguide

martinsumner:
A rewrite of the NextGenReplGUIDE as a ReplicationGuide.  Rather than a worked example, this is now an explainer of the process for configuring replication and reconciliation.

As part of this change, to avoid confusion, docs other than QuickDocs have been moved to docs/previous.  the original NextGenREPL-GettingStarted tutorial has been recovered and added to previous, so no content is lost by the rewrite.


docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+## Overview
+
+The Riak replication and reconciliation (full-sync) system has the following features:
+
+- Allows for replication between clusters with different ring sizes, `n_val`s and node counts.
+- Uses a pull model for real-time replication with low-cost queueing;
+  - Support for low-impact suspension and resumption of replication,
+  - Prevents recipient clusters from being overwhelmed by replicated PUT volumes.
+- Provides very efficient reconciliation to confirm whole clusters are synchronised;
+  - i.e. Confirmation that across multiple clusters all objects both exist, and are at the same version.
```
stephenwood24-nhs: does this line need re-wording slightly? took me a little while to understand what you are saying here

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+- [Replication triggers](#concepts---replication-triggers).
+
+### Concepts - Queues and Workers
+
+Replication queues use the [disk-backed queue framework](/docs/RiakTheoryGuide.md#disk-backed-queues) common across multiple Riak services.
+
+Replication references, which are primarily newly coordinated PUTs, are sent to the replication source queue process.  The real-time replication source then assesses which replication queues require that reference:
+
+- Each node must be configured with a separate queue for each cluster or external service that is expected to receive replication events.
+  - Each replication event will be duplicated to all queues relevant to that event, but only on one node within the cluster.
```
stephenwood24-nhs, line 73:
```suggestion
  - Each replication event will be duplicated to all queues relevant to that event, but only on one node within the source cluster.
```

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+      - Bucket name filters will work for typed and legacy buckets, with typed buckets the type will be ignored when using a name or prefix filter.
+
+For replication, the real-time replication source must be enabled on every node in the cluster, as a PUT may be coordinated from any vnode (on any node), regardless of which node received the PUT request.
+
+### Enable a Real-Time Sink
+
+There are five configuration items required to set up a sink for real-time replication.  They are all set via `riak.conf`:
+
+- `replrtq_enablesink = enabled`.
+- `replrtq_sinkqueue = <sink_cluster_name>`;
```
stephenwood24-nhs, line 166:
```suggestion
- `replrtq_sinkqueue = <sink_queue_name>`;
```

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+
+### Additional Configuration
+
+Further configuration can be added for replication using `riak.conf`:
+
+- Security configuration to enable authentication and encryption of replication traffic;
+  - `repl_cacert_filename` (required on the source);
+  - `repl_cert_filename` (required on the sink);
+  - `repl_key_filename` (required on the sink, the key associated with the certificate in `repl_cert_filename`);
+  - `repl_username` (required on the sink).
```
stephenwood24-nhs: this isn't really explained? what do these values need to be set to, what certs does it need, what's the purpose of the username?

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+- `replrtq_compressonwire`;
+  - To enable compression of replicated objects, with compression using the zlib compression library.
+  - Compression should only be enabled if objects are known to be compressible, as zlib may be computationally expensive when only limited compression can be achieved.
+- `replrtq_vnodecheck`;
+  - The `r` value on the fetch of a replication object from the source (if it has not been queued), and the `w` value on the `PUSH` of the object into the sink.
+  - The default is `one`, to ensure a minimal delay in real-time replication, but it is recommended to use the safer option of `quorum` instead.
+  - Setting to a value other than `one` will reduce the risk of mailbox overloads related to replication backlogs.
```
stephenwood24-nhs, line 216: does this only support `one` and `quorom`?

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+
+> The configuration option `tictacaae_storeheads` is not required to run all-cluster reconciliation, but is recommended to get the full operational feature set of AAE folds.
+
+When enabling Tictac AAE for the first time, it will not be usable by reconciliation until all trees have been built.  Trees will periodically rebuild, and full-sync reconciliation checks should continue to operate as expected during rebuilds.
+
+### Initial Configuration
+
+Full-sync replication requires the existence of source queue definitions and sink worker configurations, in order for discovered deltas to be repaired.  The same configurations can be used as for real-time replication.  If there is a need to support reconciliation without allowing for real-time replication - then the `block_rtq` keyword can be used instead of `any` on the source queue definition.
```
stephenwood24-nhs, line 244:
```suggestion
Full-sync replication requires the existence of source queue definitions and sink worker configurations, in order for discovered deltas to be repaired.  The same configurations can be used as for real-time replication.  If there is a need to support reconciliation without allowing for real-time replication - then the `block_rtq` keyword can be used instead of `any` as the filter on the source queue definition.
```

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+
+> In configuration, reconciliation processes are commonly referred to by the initialism `ttaaefs` - TicTac AAE Full-Sync.
+
+To enable reconciliation an initial configuration is required:
+
+- `ttaaefs_scope = all`.
+- `ttaaefs_queuename = <queue_name>`;
+  - This is the queue name to be used by the remote cluster when fetching replication events.
```
stephenwood24-nhs, line 252:
```suggestion
  - This is the local queue name to be used by the remote cluster when fetching replication events.
```

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+
+`@riak_kv_replrtq_src:handle_info:414 QueueName=replq has queue sizes p1=0 p2=0 p3=0`
+
+There is a log on each sink node of the replication timings:
+
+`Queue=~w success_count=~w error_count=~w mean_fetchtime_ms=~s mean_pushtime_ms=~s mean_repltime_ms=~s lmdin_s=~w lmdin_m=~w lmdin_h=~w lmdin_d=~w lmd_over=~w`
+
+The `mean_repltime` is a measure of the delta between the last-modified-date on the replicated object and the time the replication was completed - so this may vary if prompting replication due to real-time changes, reconciliation or seeding.  The `lmdin_<x>` counts are the counts of replicated objects which were replicated within a second, minute, hour, day or over a day.
+
+### Making runtime changes to the Sink
```
stephenwood24-nhs, line 410: What about a section on Making runtime changes to the Source ?
Registering & De-registering queues, clearing queues etc.
I think knowing how to clear queues is important in the case of a bad range-repl

docs/ReplicationGuide.md
```diff
@@ -0,0 +1,614 @@
...
+  - Users of full-sync have needed to use manually prompted rebuild windows to address this problem (i.e. a period where full-sync is suspended, and rebuilds are completed in parallel).
+  - As clusters scale, rebuild windows may become unreliable and unmanageable.
+- `riak_repl` has a keylisting form of full-sync which will do a full key and clock comparison on a vnode-by-vnode basis.
+  - A keylisting full-sync can be resource intensive and will take a significant amount of time to complete on clusters of non-trivial scale.
+
+The use of `riak_repl` is deprecated, and when a retirement schedule is agreed it will be advertised via [OpenRiak discussions](https://github.com/orgs/OpenRiak/discussions).
+
+## Replication scope
+
+Replication in Riak between clusters is limited in scope to objects only.  Cluster metadata, including bucket type configurations and authentication credentials are not replicated.  When enabling bucket types, the enablement must be triggered on each and every cluster that may receive an object for that bucket type.
```
stephenwood24-nhs: and bucket properties

stephenwood24-nhs (changes requested):

docs/ReplicationGuide.md
martinsumner: I've simplified - the exist and at the same version is duplicate information?  Just says "all objects are at the same version"

docs/ReplicationGuide.md
martinsumner: Conventions is to use cluster_name - but the name can be anything.  I've added a note in the stanza above explaining this, and another line to clarify here.

docs/ReplicationGuide.md
martinsumner: There is a third option of `all`.  I was tempted to leave undocumented as en expert feature, but I've added some lines now.

docs/ReplicationGuide.md
martinsumner: New replication-specific security section added, with link to pending security section.  Security section update awaiting approval in https://github.com/OpenRiak/riak_kv/pull/85.

stephenwood24-nhs (approved):

Issue #84, opened 2025-11-27
Grants  and Authorisation on "new" APIs


martinsumner:
Newer APIs have been added without checks for `riak_core_security` -

- AAE Fold API and Queue (fetch) API - do not check authorisation, or permission.
- The Query API - checks authorisation (but not tested), but not permission.

There are extra complications for AAE Fold and Queue, as they are used internally by `riak_kv_ttaaefs_manager`, which does not currently have support for credentials other than certificates.

martinsumner:
The Query API does have permission check - uses riak_kv.index just like 2i.

Now tested - https://github.com/OpenRiak/riak_test/pull/16

martinsumner:
Intention now is to add the authorisation check to AAE Fold, Queue and Stats requests.  However, also add a configuration option so that authorisation can be bypassed for these services when when security is enabled - so for 3.4 backwards-compatible behaviour is configurable.

However, by default (without the configuration option changing) enabling security will now disable the HTTP API and only allow for requests to all Rest endpoints via HTTPS, if they are covered by a security source.  Having an inconsistency for these endpoints by default doesn't make sense.

PR #85, opened 2025-11-28
Nhse o34 orkv.i84 querygrant

martinsumner:
Enforce authorisation checks on all HTTP APIs, but allow the previous behaviour to be supported by non-default configuration.

Document the availability of security controls in Riak Quick Docs.

Tested in https://github.com/OpenRiak/riak_test/pull/16
Resolves https://github.com/OpenRiak/riak_kv/issues/84

### DIFF
--- a/docs/OperationsAndTroubleshootingGuide.md
+++ b/docs/OperationsAndTroubleshootingGuide.md
@@ -279,9 +279,9 @@ There are three parts to Riak's security mechanism:
 
 ### TLS Enablement
 
-In Riak 3.4, support is provided for TLS 1.2 only.  The process for enabling TLS differs between the HTTP and PB interfaces.
+In Riak 3.4, support is provided for TLS 1.2 only.
 
-For HTTP, there are two steps to enabling TLS on the API:
+The process for enabling TLS differs between the HTTP and PB interfaces.  For HTTP, there are two steps to enabling TLS on the API:
 
 - Configure a listener on HTTPS within `riak.conf` - `listener.https.internal = <ip>:<port>`.
 - Configure file paths within `riak.conf` to valid [`PEM` files](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) for three components:
@@ -294,7 +294,7 @@ For HTTP, there are two steps to enabling TLS on the API:
 
 The configuration will start a HTTPS listener, and any HTTP client will be able to send any supported HTTP request via TLS using that listener.
 
-For PB, it is not possible to enable TLS in isolation without [adding further security measures](#enabling-security-and-restricting-source).  The configuration of file paths to certificate and key files is required as a prerequisite for applying those measures.  No independent listener is used for PB when security is enabled, the standard listener will expect TLS negotiation if and only if security is enabled.
+For the PB interface, it is not possible to enable TLS in isolation without [adding further security measures](#enabling-security-and-restricting-source).  The configuration of file paths to certificate and key files is required as a prerequisite for applying those measures.  No independent listener is used for PB when security is enabled, the standard listener will expect TLS negotiation if and only if security is enabled.
 
 > Riak does not support any automated certificate management, or notification on pending certificate expiry.
 
--- a/docs/OperationsAndTroubleshootingGuide.md
+++ b/docs/OperationsAndTroubleshootingGuide.md
@@ -320,16 +320,20 @@ Prior to Riak 3.4, some HTTP API requests could still be sent to the plain text 
 
 To preserve the old behaviour, and allow insecure use via HTTP of operational calls when security is enabled, the configuration option in `riak.conf` of `permit_insecure_http_ops = enabled` can be used.
 
+> Although the CLI uses the terms `user` and `password`; these would normally translate to an `application_instance` and `shared_secret` in an actual implementation.  There is no expectation that Riak security should manage the real-world usernames and passwords of operators, developers or application end-users.
+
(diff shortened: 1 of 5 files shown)


docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -262,7 +262,125 @@ Each worker pool will regularly log its current queue length and last checkout t
 
 ## Enabling Riak Security
 
-> TODO - Section on applying authentication and authorisation 
+> Riak is expected to be deployed into secure environments, it is not a database designed for direct exposure on public networks.
+
+Riak does have the optional capability to enable additional security controls, controls which are disabled by default.  However:
```
WarpEngineer:
```suggestion
Riak does have the optional capability to enable additional security controls, which are disabled by default.  However:
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -262,7 +262,125 @@ Each worker pool will regularly log its current queue length and last checkout t
...
+
+The configuration will start a HTTPS listener, and any HTTP client will be able to send any supported HTTP request via TLS using that listener.
+
+For the PB interface, it is not possible to enable TLS in isolation without [adding further security measures](#enabling-security-and-restricting-source).  The configuration of file paths to certificate and key files is required as a prerequisite for applying those measures.  No independent listener is used for PB when security is enabled, the standard listener will expect TLS negotiation if and only if security is enabled.
+
+> Riak does not support any automated certificate management, or notification on pending certificate expiry.
+
+### Enabling Security and Restricting Source
+
+To provide further security, the Riak security mechanism should be enabled. This is not possible through configuration, it must be enabled via the command line:
```
WarpEngineer:
```suggestion
To provide further security, the Riak security mechanism should be enabled. This is not possible through configuration; it must be enabled via the command line:
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -262,7 +262,125 @@ Each worker pool will regularly log its current queue length and last checkout t
...
+
+```console
+riak admin security enable
+riak admin security add-user proxy_waf
+riak-admin security add-source all 192.168.6.7/32 trust
+```
+
+This would permit access to the APIs only from the IP address `192.168.6.7` (this may be the address of a web application firewall, for example), and trust all access from that source as long as the username of `proxy_waf` is provided within the Authorization header.
+
+> In this simple case, this is functionally equivalent to applying an IP filter on the node through a standard filter utility, but it is not the security equal of that measure.  An IP filter would prevent connections being made from an unauthorised host, whereas the Risk security control allows connections and requests to be sent, but blocks requests during the processing of those requests; which presents a broader attack surface.
```
WarpEngineer:
```suggestion
> In this simple case, this is functionally equivalent to applying an IP filter on the node through a standard filter utility, but it is not the security equal of that measure.  An IP filter would prevent connections being made from an unauthorised host, whereas the Riak security control allows connections and requests to be sent, but blocks requests during the processing of those requests; which presents a broader attack surface.
```

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -262,7 +262,125 @@ Each worker pool will regularly log its current queue length and last checkout t
...
+Once security is enabled, all requests will need to have a valid `user` and a valid `source`.  There are three types of `source`:
+
+- `trust`;
+  - Applies no conditions beyond a source IP address filter;
+- `password`;
+  - Requires the user provide a valid password, as well as matching on a source IP address filter;
+- `certificate`;
+  - Supported for **the PB API only**,
+  - Requires that the username match the certificate name,
+  - By inference requires the session to have included a valid client certificate in the TLS negotiation.
```
WarpEngineer: May need to mention that the client cert must be signed by the same CA as the server cert.

Also, should we mention PAM and mark it as deprecated here?

docs/OperationsAndTroubleshootingGuide.md
```diff
@@ -262,7 +262,125 @@ Each worker pool will regularly log its current queue length and last checkout t
...
+- [`riak_kv.delete`](/docs/ObjectAPI.md#http-api-definition---delete)
+- [`riak_kv.list_keys`](/docs/OtherAPI.md#the-list-api)
+- [`riak_kv.list_buckets`](/docs/OtherAPI.md#the-list-api)
+- [`riak_kv.mapreduce`](/docs/OtherAPI.md#the-mapreduce-api)
+- `riak_kv.index`;
+  - used to control both the [legacy query api](/docs/OtherAPI.md#legacy-query-api) and the [Query API](/docs/QueryAPI.md).
+
+For all other API endpoints, only `source` protection is applied.
+
+> With the PB API, authentication is provided at the start of a connection, and grants are assessed and cached for that connection to be used against each request.  One the HTTP API, each request on a connection is authenticated and has grant checks made independently on a per-request basis.
```
WarpEngineer:
```suggestion
> With the PB API, authentication is provided at the start of a connection, and grants are assessed and cached for that connection to be used against each request.  On the HTTP API, each request on a connection is authenticated and has grant checks made independently on a per-request basis.
```

WarpEngineer (changes requested):

docs/OperationsAndTroubleshootingGuide.md
martinsumner: Updated - https://github.com/OpenRiak/riak_kv/pull/85/commits/e091db55d7286ad0cfea32730942e2e9dfd2b5d4

WarpEngineer (approved):

Issue #102, opened 2025-12-16
Query API and compiled Regex


martinsumner:
The Query API can be passed regular expressions in two places:

- A regular expression as an alternative to the use of a filter and eval expression;
- A regular expression as part of an eval expression to extract projected attributes.

In both cases the regular expressions are compiled on receipt of the query:

- The `riak_kv_query` module compiles regular expressions to be distributed (when it is an alternative to filter/eval expressions);
- When the eval parser code is applied, it compiles any eval regex function (when it is part of an eval expression).

This works .... normally.

However - https://github.com/erlang/otp/pull/9976.

There will be circumstances, perhaps during a rolling hardware upgrade, when this might fail.  It will outright fail when compiling with OTP 28 which makes the compiled regular expression an unportable reference.  In OTP 28, via the PR above the export/import capability is provided.

This needs to be fixed to prevent unexpected errors, but also in such a way that migration to OTP 28 is not impaired - especially with node-by-node migration.  i.e. so that an OTP26/3.4 node does not distribute a compiled regex to a OTP28/4.0 node that cannot process the compiled expression.

For regular expressions embedded within evaluation expressions, there seems to be no way of handling this that supports compilation.  As the regular expression exists within an anonymous function, there would be no way to recompile that function after passing it to another node.  So it seems this must be changed prior to the release of 3.4 to not support pre-compilation.

For regular expressions passed as alternative to eval/filter expression; then perhaps there should be a switch to turn central compilation on/off?  Or perhaps it should be off until Riak 4.0.

If it is not compiled, should it be compiled at each vnode?  If there are no range matches, the compilation is unnecessary.  The relative difference with a simple regular expression is still small with < 100 terms per vnode in the range.

With OTP 28 it will be possible to pattern match as regular expression, and if it is not IOData, assume it is compiled and use `re:import/1` to avoid re-compiling each vnode.

Perhaps it would be best to send a tuple of {IORE, CompileRE}, and in the current code {IORE, none}.  So that a vnode running pre-4.0 will always check just the first element of the tuple, but a vnode running >=4.0 and >=28.1 can inspect the second element of the tuple and import if it is not `none`. 

If an uncompiled regular expression is passed, the value of compiling it multiple times (once per vnode) is sufficiently uncertain to make it unjustified.

PR #103, opened 2025-12-16
Don't use compiled regex unless on local node

martinsumner:
Prepare the groundwork for non-local exported regex to be available in OTP 28.

Fixes the `riak_kv_backend:complex_query/7` spec (which was incorrect, the spec is not changed by the PR)

### DIFF
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -28,14 +28,10 @@
     :: fun((binary(), any()) -> any() | no_return()).
 -type fold_keys_fun()
     :: 
-        fun(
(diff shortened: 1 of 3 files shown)


martinsumner:
https://github.com/OpenRiak/riak_kv/issues/102

Also see - https://github.com/martinsumner/leveled/pull/484

martinsumner:
Note that documentation had assumed they were not compiled, so no doc change required:

https://github.com/OpenRiak/riak_kv/blob/d8541d54e2c211a5d2082eeb5d6a20ffd581060a/docs/QueryAPI.md?plain=1#L78

ThomasArts (approved):

Issue #88, opened 2025-12-02
Intermittent eunit failure - `riak_kv_ttaaefs_manager:take_first_workitem_test/0`


martinsumner:
There is an intermittent eunit test failure with a `riak_kv_ttaaefs_manager` test:

```erlang
riak_kv_ttaaefs_manager: take_first_workitem_test...*failed*
in function riak_kv_ttaaefs_manager:take_first_workitem_test/0 (/home/runner/work/riak_kv/riak_kv/src/riak_kv_ttaaefs_manager.erl, line 1655)
in call from eunit_test:'-mf_wrapper/2-fun-0-'/2 (eunit_test.erl, line 273)
in call from eunit_test:run_testfun/1 (eunit_test.erl, line 71)
in call from eunit_proc:run_test/1 (eunit_proc.erl, line 531)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 356)
in call from eunit_proc:handle_test/2 (eunit_proc.erl, line 514)
in call from eunit_proc:tests_inorder/3 (eunit_proc.erl, line 456)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 346)
**error:{assertMatch,[{module,riak_kv_ttaaefs_manager},
              {line,1655},
              {expression,"PromptS3YetMoreSeconds > PromptS2YetMoreSeconds"},
              {pattern,"true"},
              {value,false}]}
  output:<<"">>
```

martinsumner:
This is as a result of the `?SCHEDULE_JITTER`:

https://github.com/OpenRiak/riak_kv/blob/f4ed77e0c17caf14fe70d12a3b3481b0181f8cc5/src/riak_kv_ttaaefs_manager.erl#L80-L81

The schedule is determined by dividing up the 24 hour day by the number of work items, then by the number of nodes (so that each node doesn't overlap) and then by the cluster_slice (so that same nodes on different clusters don't coordinate) - and then the `?SCHEDULE_JITTER` is added.

In the unit test, there are 48 work items - and so 30 minutes to schedule the work.  There are then 8 nodes, and 4 cluster slices - and so each cluster slice and node is just less than 1 minute apart.  The jitter may differ by 59s, so ordering is not guaranteed.  ClusterSlice N + 1 may be triggered before ClusterSlice N.

[...]

PR #105, opened 2025-12-29
Don't jitter by more than the available time slice

martinsumner:
The jitter randomises spacing, but shouldn't cause out of order scheduling.

### DIFF
--- a/src/riak_kv_ttaaefs_manager.erl
+++ b/src/riak_kv_ttaaefs_manager.erl
@@ -77,6 +77,7 @@
     % Reset the range if the number of results for repair are less than
     % MaxResults div ?RANGE_DIVISOR.  Either all the results have been
     % returned, or the range has narrowed and is no longer effective
+
 -define(SCHEDULE_JITTER, 60).
     % Jitter each schedule start time by up to a minute, to avoid overlaps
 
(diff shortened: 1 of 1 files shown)


martinsumner:
https://github.com/OpenRiak/riak_kv/issues/88

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -1399,25 +1400,31 @@ take_next_workitem([NextAlloc|T], Wants,
             SliceNumber,
             ClusterSlice,
             NodeCount,
-            SliceCount),
+            SliceCount
+        ),
     {Mega, Sec, _Micro} = ScheduleStartTime,
+    Slices = lists:sum(lists:map(fun(W) -> element(2, W) end, Wants)),
```
ThomasArts: Since you now have types, we see that `schedule_want` always is a 2-tuple. I would therefore prefer to see the mapping in this function:
```suggestion
    Slices = lists:sum(lists:map(fun({_, WItem}) -> WItem end, Wants)),
```
At least, if the type system is happy with that.

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -1399,25 +1400,31 @@ take_next_workitem([NextAlloc|T], Wants,
...
+    MinSpacing = max(1, ?SECONDS_IN_DAY div (Slices * NodeCount * 4)),
     ScheduleSeconds =
         Mega * ?MEGA
-     + Sec + SecsFromStartTime
-     + rand:uniform(?SCHEDULE_JITTER),
+        + Sec
+        + SecsFromStartTime
+        + rand:uniform(min(?SCHEDULE_JITTER, MinSpacing)) - 1,
```
ThomasArts, line 1412: You take MinSpacing minimum as 1, because otherwise `rand:uniform` does not work. But you want it to possibly be 0 Schedule Seconds between and hence a `-1`. As a consequence you now change the behaviour, because for all "normal" cases, SCHEDULE_JITTER is still 60 and will now add between 0 and 59 seconds, whereas before between 1 and 60.
This is probably fine, but the `-1` doesn't really contribute much, I would say.

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -1399,25 +1400,31 @@ take_next_workitem([NextAlloc|T], Wants,
...
     {MegaNow, SecNow, _MicroNow} = os:timestamp(),
     NowSeconds = MegaNow * ?MEGA + SecNow,
     case ScheduleSeconds > NowSeconds of
         true ->
             {NextAction, ScheduleSeconds - NowSeconds, T, ScheduleStartTime};
         false ->
-            ?LOG_INFO("Tictac AAE skipping action ~w as manager running "
-                        ++ "~w seconds late",
-                        [NextAction, NowSeconds - ScheduleSeconds]),
+            ?LOG_INFO(
```
ThomasArts, line 1419: Statistically this is going to happen slightly more often now that you subtract 1 second from schedule seconds

ThomasArts (approved):

src/riak_kv_ttaaefs_manager.erl
martinsumner: I agree that it isn't contributing much.  I still prefer 0.59 as + 60, I would argue, is in the next time slice (although as the next time slice will be in 1..60 they wouldn't overlap).

Point taken but I'l leave unchanged.

src/riak_kv_ttaaefs_manager.erl
martinsumner: Yes, but we're applying a brake in this case due to the possibility of an issue - so it is safe to err on the side of a bit more often.

Issue #69, opened 2025-11-02
Property-test failure - `replrtq_snk_eqc`


martinsumner:
Testing on OTP 26 with testing_budget = 600:

```
replrtq_snk_eqc:config(enabled,
    #{queue => queue2, workers => 20,
      peers => [{{"history.org", 8012, http}, {active, 13}}],
      peerlimit => 20}) ->
  ok.
replrtq_snk_eqc:start() -> <0.1249636.0>.
replrtq_snk_eqc:remove(queue2) -> ok.
replrtq_snk_eqc:sleep(6) -> ok.
replrtq_snk_eqc:add(#{queue => sledge_hammer, workers => 22,
      peers
        => [{{"permitted.com", 8006, http}, {active, 13}},
            {{"pleasure.org", 8000, http}, {active, 10}},
            {{"units.org", 8008, http}, {active, 5}},
            {{"members.com", 8002, http}, {active, 10}},
            {{"inner.org", 8002, http}, {active, 5}},
            {{"sm.de", 8000, http}, {active, 7}},
            {{"bodies.de", 8000, http}, {inactive, 8}},
            {{"conversion.com", 8005, http}, {active, 18}}],
      peerlimit => 11}) ->
  ok.
replrtq_snk_eqc:add(#{queue => default, workers => 16,
      peers => [{{"ebony.org", 8023, http}, {active, 5}}],
      peerlimit => 16}) ->
  ok.
replrtq_snk_eqc:add(#{queue => c, workers => 18,
      peers
        => [{{"yards.com", 8003, http}, {error, 17}},
            {{"shirts.org", 8000, http}, {error, 5}},
            {{"members.co.uk", 8004, http}, {active, 8}},
            {{"cocks.de", 8030, http}, {active, 5}},
            {{"emails.org", 8027, http}, {active, 10}},
            {{"stick.edu", 8033, http}, {active, 10}},
            {{"php.co.uk", 8012, http}, {active, 13}},
            {{"last.edu", 8025, http}, {active, 18}}],
      peerlimit => 4}) ->
  ok.
replrtq_snk_eqc:remove(sledge_hammer) -> ok.
replrtq_snk_eqc:sleep(8) -> ok.
replrtq_snk_eqc:resume(queue2) -> not_found.
replrtq_snk_eqc:add(#{queue => sledge_hammer, workers => 4,
      peers
        => [{{"central.com", 8001, http}, {active, 5}},
[...]

PR #106, opened 2025-12-30
Nhse o34 orkv.i69 eqc fix

martinsumner:
Resolves issues with test, now passing with 5000 tests run.

Corrects the max count in two scenarios.  See https://github.com/OpenRiak/riak_kv/issues/69 for more detailed explanation of test failures.


ThomasArts (approved):

eqc/replrtq_snk_monitor.erl
```diff
@@ -27,6 +27,7 @@ start_link() ->
     gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
 stop() ->
+    timer:sleep(1), % Wait 1ms for any spawned fetches to arrive
```
ThomasArts: I verified that instead of sleep, you could also do an `erlang:yield()` here, but essentially the same thing.

martinsumner: I agree that yield() is better.  Will change.

Issue #111, opened 2026-02-09
Remove error_logger


martinsumner:
Merge in @tburghart changes to remove the error_logger, and confirm no impact on eqc tests.

martinsumner:
https://github.com/OpenRiak/riak_core/pull/38
https://github.com/OpenRiak/riak_kv/pull/112

PR #112, opened 2026-02-09
Nhse o34 orkv.i111 rmverrorlogger

martinsumner:
wday error_logger removal.

Confirmed to pass all eqc tests

### DIFF
--- a/eqc/crdt_statem_eqc.erl
+++ b/eqc/crdt_statem_eqc.erl
@@ -1,8 +1,7 @@
+%% -*- mode: erlang; erlang-indent-level: 4; indent-tabs-mode: nil -*-
 %% -------------------------------------------------------------------
 %%
-%% crdt_statem_eqc: Quickcheck statem test for riak_dt modules
-%%
-%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2013-2016 Basho Technologies, Inc.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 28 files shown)


martinsumner:
https://github.com/OpenRiak/riak_kv/issues/111

martinsumner:
Depends on https://github.com/OpenRiak/riak_core/pull/38 - before reverting `riak_core` branch

martinsumner:
Currently awaiting decision on `riak_kv_env` changes.  Refactoring of system calls hits an issue on Ubuntu, where `usr/bin/ulimit` is not a file (`ulimit` is just a shell script).

martinsumner:
There are a couple of caveats here with regards to `riak_kv_env:doc_env/0`.

The settings tested don't appear to align with the recommendations in the reference (e.g. `netdev_max_backlog`, `somaxconn`): https://fasterdata.es.net/host-tuning/linux/.  I think the general recommendation should be to trust defaults, as mentioned in the reference, as it is not possible to know how Riak is to be used, recommending retuning without context may be harmful.  We don't know if Riak is going to be on a 1Gbps or 100Gbps NIC, what the network delay will be etc, so how can we recommend buffer sizes?

There's some out-of-date checks as well:

- The ETS Max tables check is no longer relevant; https://www.erlang.org/doc/apps/stdlib/ets.html.
- The async processes defaults to 1 now that dirty NIFs are used, and I think that should be OK; https://www.erlang.org/doc/apps/erts/erl_cmd.html#emulator-flags.
- It might be worth guarding against people disabling `net.ipv4.tcp_timestamps` (it is often an erroneous recommendation of pen test reports) as this will disable RFC 6191 support; https://www.rfc-editor.org/rfc/rfc6191.html.
- The ulimit of 32K is much too low for leveled.

There is no testing of any of this within the OpenRiak community, and no capacity to run tests.

For QuickDocs, I took the approach approach to deliberately not include settings for OS tuning as it is hard to maintain - how do we know they are still relevant as the kernels evolve?

Also, I think, if we recommend VM settings changes, then it should be done with a reference to the cuttlefish option to change it.

tburghart:
@martinsumner  I have a revision in progress that addresses the above, should have it pushed today.

tburghart:
@martinsumner Branch `contrib/trb/riak_kv_env-tuneup` (from your branch) contains a tune-up of the `riak_kv_env` module that, among other things, points users to where they can change the setting.
Note that the Cuttlefish keys are from my WiP branch at https://github.com/OpenRiak/cuttlefish/blob/wday/trb/schema-wip/priv/erlang_vm.schema - some of these _may_ differ from, or be in addition to, what's currently in the Riak schema.

* I've left the ETS table limit in with an explanation as to why.
* I've changed the behavior of the async threads threshold to log a message _only_ if it's high (suggesting a leftover setting from Riak v2).
* I've added a `net.ipv4.tcp_timestamps` check on Linux, probably should look into it on \*BSD as well, but not today.
* I've left most other network tuning as it was - you're correct that we can't possibly recommend the best config for everyone, but maybe we can get their attention if they've ignored it completely.

WarpEngineer (approved):

PR #121, opened 2026-02-19
Use metadata_fetch not dict for vtag check

martinsumner:
An example of the use of dict() missed when updating for vi metadata


martinsumner:
Test - https://github.com/OpenRiak/riak_test/pull/22

WarpEngineer (approved):

PR #122, opened 2026-02-22
riak admin vnode status in JSON format, clique-conformant

hmmr:
`riak admin vnode-status | sed -n 1p | json_pp` will now produce vnode backend status (for leveled, collected by `leveled_bookie:book_status/1`) as a JSON blob, like this:

```
{
   "dev1@127.0.0.1" : [
      {
         "backend" : "riak_kv_leveled_backend",
         "backend_status" : {
            "fetch_count_by_level" : {
               "0" : {
                  "count" : 0,
                  "time" : 0
               }, 
...
            },
            "get_body_time" : 0,
            "get_sample_count" : 0,
            "head_rsp_time" : 0,
            "head_sample_count" : 0,
            "journal_last_compaction_duration" : "undefined",
            "journal_last_compaction_max" : "undefined",
            "journal_last_compaction_mean" : "undefined",
...
            "put_ink_time" : 0,
            "put_mem_time" : 0,
            "put_prep_time" : 0,
            "put_sample_count" : 0
         },
         "counter" : 210000,
         "counter_lease" : 220000,
         "counter_lease_size" : 10000,
         "counter_leasing" : false,
         "idx" : "1210306043414653979137426502093171875652569137152",
         "vnodeid" : "0xe42f1fd9312f8641"
      },
      {
...
```
The command accepts `-n NODE|all` option, to collect and print the status from a specified NODE, or all nodes.

See https://github.com/basho/riak_kv/issues/343.

### DIFF
--- a/src/riak_kv_vnode_status_cli.erl
+++ b/src/riak_kv_vnode_status_cli.erl
@@ -25,7 +25,6 @@
 -include_lib("kernel/include/logger.hrl").
 
 -export([register_cli/0]).
--export([vnode_status_on_nodes/2, tableify/1, tableify1/1]).
 
 register_cli() ->
     register_all_usage(),
@@ -61,7 +60,7 @@ get_vnode_status_specs() ->
 
 get_vnode_status_cmd([_, _ | Args], _, Options) ->
     Nodes = extract_nodes(Options),
-    PerNode = [ {Node, [[{idx, integer_to_binary(Idx)} | tableify(X)] || {Idx, X} <- Res]}
+    PerNode = [{Node, [{integer_to_binary(Idx), jsonify1(X)} || {Idx, X} <- Res]}
                 || {Res, Node} <- vnode_status_on_nodes(Nodes, [])],
     case Args of
         [] ->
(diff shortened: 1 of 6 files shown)


martinsumner:
This doesn't work on all backends - e.g. memory backend.  Perhaps just make sure that this is handled neatly without an RPC error i.e. we should still get the vnode status information, if not the backend.  Haven't tested with multi-backend, but this might also be an issue there.

The json output is a bit busy.  Piping into a pretty printer might give issues with the idx being an integer (i.e. it is too big an integer and so will be corrupted).  Perhaps the idx should be a string to avoid this.

It would be good to have an -i switch to request a specific partition.

hmmr:
Did all backends including multi and multi_prefix, and also added option -p (example below is from a cluster of dev1 and dev2, with different backends):

```
$ ./dev/dev1/riak/bin/riak admin vnode-status -p 0 -n all  | sed -n 1p | json_pp
Duplicate specification "V" for option "v"
{
   "dev1@127.0.0.1" : {
      "0" : {
         "backend" : "riak_kv_bitcask_backend",
         "backend_status" : {
            "key_count" : 0,
            "status" : []
         }, 
         "counter" : 920000,
         "counter_lease" : 930000,
         "counter_lease_size" : 10000,
         "counter_leasing" : false,
         "vnodeid" : "0xe42f1fd9312f8611"
      }  
   }, 
   "dev2@127.0.0.1" : {
      "0" : {
         "backend" : "riak_kv_eleveldb_backend",
         "backend_status" : {
            "compactions" : 0,
            "files_size_mb" : 0,
            "fixed_indexes" : true,
            "level" : 1,
            "read_block_error" : "0",
            "read_mb" : 0,
            "time" : 0,
            "write_mb" : 0
         }, 
         "counter" : 230000,
         "counter_lease" : 240000,
         "counter_lease_size" : 10000,
         "counter_leasing" : false,
         "vnodeid" : "0x7dcd79d831331f50"
      }  
   }  
}  
```

Test is updated, too.

src/riak_kv_vnode_status_cli.erl
```diff
@@ -0,0 +1,194 @@
...
+
+
+get_vnode_status_cmd([_, _ | Args], _, Options) ->
+    Nodes = extract_nodes(Options),
+    Partitions = extract_vnodes(Options),
+    PerNode = [{Node, [{integer_to_binary(Idx), jsonify1(X)} || {Idx, X} <- Res]}
+                || {Res, Node} <- vnode_status_on_nodes(Nodes, Partitions, [])],
+    case Args of
+        [] ->
+            [clique_status:text(mochijson2:encode(PerNode))];
```
martinsumner: When new encode/decode functions into riak_kv - please use `riak_kv_wm_json` (which is based on the OTP JSON) not `mochijson2`.

In `riak_test` it is OK to continue to use mochijson2.

martinsumner:
I think there should be a sub-section added to QuickDocs, in the "Logging and Statistics" section following on from https://openriak.github.io/riak_kv/OperationsAndTroubleshootingGuide.html#riak-stats

```
### Vnode Status
{: .d-inline-block }

Available from Riak 3.4.1
{: .label .label-purple }

A significant proportion of the work within Riak takes places within the vnode.  To see the status of each vnode in the cluster, and see available statistics from the backend:  `riak admin vnode-status | sed -n 1p | json_pp`

To look at the statistics from specific nodes or partitions see: `riak admin vnode-status --help`.
```

I did think it would be better if the console command pretty-printed the JSON for us.  The `riak_kv_wm_json` is a copy of the json module from OTP, but I took it from OTP 27.0, not OTP 27.1 which added the `format/1` alternative to `encode/1`.

Ideally we should update `riak_kv_wm_json` to the 27.1 version and pretty-print.  However, I think having the advice in QuickDocs to use the ` | sed -n 1p | json_pp` pipeline is good enough.  With the docs update I'm happy to approve, the json update isn't necessary.

martinsumner:
One additional question - what happens during a transfer?  Where multiple nodes have the same vnode.  I don't expect it to have any issues, but it is worth validating that.

Is it just primary vnodes that are shown?  The QuickDocs update should just clarify what the operator will and won't see in handoff/transfer scenarios.

hmmr:
> One additional question - what happens during a transfer? Where multiple nodes have the same vnode. I don't expect it to have any issues, but it is worth validating that.
> 
> Is it just primary vnodes that are shown? The QuickDocs update should just clarify what the operator will and won't see in handoff/transfer scenarios.

During transfers, partitions being transferred will appear under both nodes. Which vnode is sending and which is receiving, is not indicated.

martinsumner (approved):

PR #123, opened 2026-02-27
Add `queue_raw_keys` and `queue_raw_terms`

martinsumner:
It may be possible to require a large `raw_keys` or `raw_terms` query that may require excessive amounts of memory to accumulate.

In this case, the results can now be queued (on disk), and the fetched in batches from the queue (potentially by multiple processes).

This requires two new accumulation_options `queue_raw_keys`/`queue_raw_terms`.  The Queue HTTP API now supports a `GET` as well as a `POST` so that results can be fetched, (async via `GET` following a `POST`).

### DIFF
--- a/src/riak_kv_query_filebuffer.erl
+++ b/src/riak_kv_query_filebuffer.erl
@@ -175,7 +175,11 @@ return_results({Node, Pid, Reference}, MaxResults, B) ->
                     [UnexpectedError]
                 ),
                 {error, unexpected_error}
-    end.
+    end;
+return_results(_UnexpectedRefFormat, _MaxResults, _B) ->
+    ?LOG_WARNING("Queue reference received in unexpected format"),
+    {error, unexpected_error}.
+
 
 -spec query_complete(pid()) -> {ok, non_neg_integer()}.
 query_complete(Pid) ->
--- a/src/riak_kv_query_filebuffer.erl
+++ b/src/riak_kv_query_filebuffer.erl
@@ -91,8 +91,8 @@
     }
 ).
 
--define(RSPCOUNT_KEY, <<"responses_count">>).
--define(RCVCOUNT_KEY, <<"received_count">>).
+-define(RSPCOUNT_KEY, <<"returned_count">>).
+-define(RCVCOUNT_KEY, <<"queued_count">>).
 -define(QCOMPLETE_KEY, <<"query_complete">>).
 
 -type inbound_results() ::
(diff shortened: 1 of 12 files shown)


martinsumner:
WIP - pending QuickDoc update

src/riak_kv_query_server.erl
```diff
@@ -204,6 +207,18 @@ init(Query) ->
                         #list_acc{};
                     raw_keys ->
                         #list_acc{};
+                    queue_raw_keys ->
```
ThomasArts: possibly a helper function for this chunk? It's two times the same code, with only difference last element in list that is second argument to `riak_kv_query_filebuffer_sup:start_query_filebuffer`

ThomasArts: `raw_query_acc(Query, Bucket, Kind)` or so

src/riak_kv_query_filebuffer_sup.erl
```diff
@@ -0,0 +1,57 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc supervise the riak_kv query servers used to
```
ThomasArts: Please document why this is a supervisor, since it is not in order to restart failing children (but to know that a certain pid is a filebuffer).

src/riak_kv_query_filebuffer.erl
```diff
@@ -0,0 +1,654 @@
...
+-ifdef(TEST).
+
+check_pid(Pid) -> is_process_alive(Pid).
+
+-else.
+-spec check_pid(pid()) -> boolean().
+check_pid(Pid) ->
+    is_process_alive(Pid) andalso
+        not
+            ({error, not_found} ==
```
ThomasArts: This is where the supervisor behaviour is used.
One could also drop the `not`  and write
            `{error, not_found} /= ...`
Saves some characters

src/riak_kv_query_filebuffer.erl
```diff
@@ -0,0 +1,654 @@
...
+            queue_reference = QueueRef,
+            accumulation_option = AccOpt,
+            bucket = Bucket
+        },
+        InactivityTimeoutMS
+    }.
+
+handle_cast(
+    {aggregate_results, Results}, State)
+        when not State#state.query_complete ->
```
ThomasArts: Possibly just not have this guard here but inside the body and return `State` in the unlikely case the guard does not hold.

ThomasArts (approved):

src/riak_kv_query_server.erl
martinsumner: Creating an init_acc(AccType) function to replace the long series of case statements to return the Acc.

ThomasArts (approved):
More robust!

ThomasArts (approved):

PR #144, opened 2026-04-24
Nhse o40 orkv.i141 silvermachine

martinsumner:
Switch the HTTP aPI to use SilverMachine (within riak_api) not mochiweb/webmachine.

See https://github.com/OpenRiak/riak_api/pull/3


WarpEngineer (approved):
I tried to find a problem but the logic seem sound here. It's a very large PR so possible I missed something. We'll need a good bit of testing.

PR #146, opened 2026-05-06
Use default max_results if not provided

martinsumner:
Previously there was a bug whereby not providing max_results would crash the request

### DIFF
--- a/src/riak_kv_wm_query.erl
+++ b/src/riak_kv_wm_query.erl
@@ -320,30 +320,23 @@ malformed_request(RD, Ctx) when Ctx#ctx.method =:= 'GET' ->
     BT = riak_kv_wm_utils:maybe_bucket_type(Ctx#ctx.bucket_type, Bucket),
     case wrq:get_qs_value("result_queue", RD) of
         QueueString when is_list(QueueString) ->
+            QueueBin = list_to_binary(QueueString),
             case wrq:get_qs_value("max_results", RD) of
                 undefined ->
-                    application:get_env(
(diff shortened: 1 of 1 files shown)


src/riak_kv_wm_query.erl
```diff
@@ -320,30 +320,23 @@ malformed_request(RD, Ctx) when Ctx#ctx.method =:= 'GET' ->
     BT = riak_kv_wm_utils:maybe_bucket_type(Ctx#ctx.bucket_type, Bucket),
     case wrq:get_qs_value("result_queue", RD) of
         QueueString when is_list(QueueString) ->
+            QueueBin = list_to_binary(QueueString),
             case wrq:get_qs_value("max_results", RD) of
```
hmmr, line 324: There is a convenient `wrq:get_qs_value/3` (the one with a default value if no such parameter is in `RD`), which would eliminate a case clause. I'm not insisting on this, however, since this entire module will be gone.

I note that it's already done in https://github.com/OpenRiak/riak_kv/pull/144/changes#diff-c5e568236b57f36df28011b58ecce5d0bfb6d28c1054a4e1a1b5c604f246ae1bR202.

hmmr (approved):

src/riak_kv_wm_query.erl
martinsumner: I'll stick with it as-is, due to issues with handling the type of value (i.e. the environment variable being an integer, and the qs_value being a string), so it won't be much simpler.

Issue #151, opened 2026-05-27
Issues with key amnesia and full-sync


martinsumner:
Background:

- Two clusters Active/Passive, Riak database with o(100 billion) objects, using nextgenrepl.
- Database has a bucket with o(10 billion) tombstones
- Attempt is made to reap tombstones on mass
- .... about 90% of reaps work fine on both sites, but something goes wrong, about 10% of reaps occur in the active site from which the reap was triggered but not on the passive site
  - working assumption that this is related to https://github.com/OpenRiak/riak_kv/issues/49
- ... failure goes unnoticed there are now about 1 billion deltas between the clusters, where a tombstone exists on one site but not another.

Resolving this through full-sync is hampered, in part due to the scale of the data-set (and the delta), but also due to key amnesia.

Due to handling of key amnesia, when full-sync detects a tombstone on Passive, that is not present on Active (as the reap worked on Active) - the tombstone is replicated from Passive to Active resolve.  However, on Active key amnesia is triggered (as on the vnodes in the preflist are in the vector clock , but the object is no longer present).  Key amnesia causes the write to have its clock incremented on Active (first locally and then via a triggered read repair to spread the resolution intra-cluster).  However, now there is still a delta - both Active and Passive clusters now have the tombstone but Active has a more advanced vector clock.

So the issue is only fixed after full-sync detects the new discrepancy and re-replicates the  tombstone from Active to Passive.

There is a kicker to this though.  Optimisations to full-sync are about detecting the range (in terms of last-modified date), and using that ranges on subsequent full-sync operations to speed up Key/Clocks queries (which on a database this size would otherwise take hours to complete).

This normally helps as replication incidents tend to impact objects that were modified in a specific time range.  However in this case:

- The last-modified date of the object is unrelated to the timing of the fault (the unreaped tombstone could have been created any time before the failure to replicate the reap operation).
- The key amnesia event updates the vector clock (and the timestamp in the vector clock), but not the last-modified date for the object.  The full-sync manager uses the timestamp in the vector clock to determine the modified date range, but the query in the next sync operation uses the modified date on the object.  So full-sync thinks there was an issue when the amnesia was resolved, but then doesn't detect any out-of-sync keys in that range due to this mismatch of how ranges are set and applied.

martinsumner:
The workaround is to suspend auto full-sync, and trigger specific per-bucket ysnc operations with a operator-defined range.  Once the delta is closed, full-sync will then be re-enabled.

[...]

PR #153, opened 2026-06-04
Add non-standard option to replicate repaired tombs

martinsumner:
When a tomb-stone is resurrected through replication, it will commonly prompt a clock change due to key amnesia.

After the change has been made locally, a read repair event is triggered.  this configuration option allows the repaired object to be auto-replicated.  In the case of a resurrected tombstone, it will now have an updated clock, and so will otherwise be another delta to be resolved via full-sync.

Draft PR pending riak_test


martinsumner:

martinsumner:
In  the riak_test, the behaviour with the configuration disabled was actually better.  The need to r=all the reader entries slowed down the internal repair - a reader backlog developed.  This led to more repeated work.

Although everything was replicated twice without the configuration - there were no backlogs, and it was just twice.  Whereas with backlogs things were replicated multiple times

martinsumner:
Overall, I don't think this PR should be added.  It would require non-standard intervention (due tot he risk of perhaps having rebounding replication if enabled by default), but once non-standard intervention is required, there is probably better intervention to make e.g. bucket-sync.

PR #154, opened 2026-06-04
Make TreeSize in Range configurable

martinsumner:
A larger tree size will be slower in merging the trees, but will allow a more granular view of keys/clocks.

As well as making the key size configurable, a helper function to resync a bucket is added.  This will resync using a loop of individual exchanges, but breaking up the key-space using segment filters.  This, in testing, is a fast way to resync an individual bucket - handling the situation when the delta space is o(tree-size) much better than leaving full-sync to slowly work through.

The helper function uses some sane defaults, but the underlying function allows for a flexible configurable approach.

This is based on some scripting done for a large Riak user that had a o(1bn) delta in a o(10bn) size bucket after some  failed operational actions.  In this case KeySpace ranges were used, but for buckets of smaller sizes this should be unnecessary - the use of segment space should be sufficient.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -73,6 +73,7 @@ ttaaefs
 aaefold
 quickdocs
 keyname
+resync
 
 # Riak configuration terms
 cacert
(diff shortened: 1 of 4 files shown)


martinsumner:

martinsumner:
https://github.com/OpenRiak/riak_test/pull/30

martinsumner:
To be re-opened with a better description

PR #156, opened 2026-06-11
CLI support for riak_client:repair_node()

hmmr:
Expose `riak_client:repair_node/0` as `riak admin node repair`.

Related PRs: https://github.com/OpenRiak/riak/pull/36. Test in https://github.com/OpenRiak/riak_test/pull/29.

### DIFF
--- a/src/riak_kv_node_cli.erl
+++ b/src/riak_kv_node_cli.erl
@@ -140,20 +140,19 @@ node_repair_status_cmd(_Cmd, _Args, Opts) ->
 get_node_repair_status(Nodes) ->
     [begin
          Vnodes = erpc:call(Node, riak_core_vnode_manager, all_vnodes, []),
-         Statuses = [{Mod, Idx, Pid,
+         Statuses = [{Mod, Idx,
                       erpc:call(Node, riak_core_vnode_manager, repair_status, [{Mod, Idx}])}
-                     || {Mod, Idx, Pid} <- Vnodes],
-         {Node, [{Mod, Idx, Pid} || {Mod, Idx, Pid, Status} <- Statuses, Status /= not_found]}
+                     || {Mod, Idx, _Pid} <- Vnodes],
+         {Node, [{Mod, Idx} || {Mod, Idx, Status} <- Statuses, Status /= not_found]}
      end || Node <- Nodes].
 
 nodes_running_repair() ->
     AllSS = get_node_repair_status(get_nodes()),
     [N || {N, SS} <- AllSS, SS /= []].
 
-jsonify_status({Mod, Idx, Pid}) ->
+jsonify_status({Mod, Idx}) ->
     #{mod => Mod,
-      idx => Idx,
-      pid => list_to_binary(pid_to_list(Pid))}.
+      idx => Idx}.
 
 node_repair_start_cmd(_Cmd, _Args, Opts) ->
     Node =
--- a/src/riak_kv_node_cli.erl
+++ b/src/riak_kv_node_cli.erl
@@ -64,7 +64,7 @@ node_repair_start_usage() ->
 
 node_repair_stop_usage() ->
     ["riak admin node repair stop [-n|--node NODE|all] REASON\n",
-     "Kill all an ongoing partition repair on NODE, with REASON.\n"
+     "Kill all ongoing partition repair on NODE, with REASON.\n"
     ].
 
 -define(NODEOPT, {node, [{shortname, "n"},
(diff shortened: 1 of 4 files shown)


martinsumner:
I'm not sure about repairing multiple nodes, what happens when partitions are covered in more than one nodes - the behaviour might be unpredictable.  Having nodes of `all` as well, I'm not sure it is obvious what will happen in this case.

I've only done non-functional testing for repairing a single node, so I think it should be restricted to repairing a single node at a time.  There are are advanced configurable options like `double-pair` that might not behave as expected with multiple node repairs being triggered.

hmmr:
Would it, then, be appropriate to extend the action of `riak admin node repair` with `status` subcommand which would report if any repairs are in progress? And if so, prevent a new repair from starting?

martinsumner:
Repairs can be tracked like other handoffs/transfer operations - so there's not a need for a status, in that there is another way of following this .

However, there is a repair_status for each vnode (tracked by the vnode_manager).  So just as node_repair is just a wrapper to call repair on all vnodes on the node, there could be a call to repair_status that collates the status across vnodes - but the status is just in_progress/not_found.

There are better ways to track progress - so I'm not sure it is necessary, but it could give you a definitive in_progress/complete statement.

There may be more complications here.  I can't remember the history behind this comment:

https://github.com/OpenRiak/riak_test/blob/a9605720be2efa567dde65489b0caf415f7e4d92/tests/node_repair_big.erl#L187

Perhaps I hadn't thought of using repair_status, or perhaps there was an issue with it.

hmmr:
I have added subcommands `repair status` and `repair stop`, and extended `start` with a check to only allow repair to start when no other repairs are running.

martinsumner:
Could the new commends be added to the riak_test suite in some way.  Perhaps the status command could be used to track repairs in https://github.com/OpenRiak/riak_test/blob/openriak-3.4/tests/node_repair_nval.erl.

Should there be a test that after start then stop - it is possible to start a repair again, and have it complete?

hmmr:
Test in https://github.com/OpenRiak/riak_test/pull/29 updated to:

* check that only one repair job can run at any time on any node;
* include stop and resume, checking that pre-stop and post-resume states match;
* test json output;
* use wait_until_repairs_complete/1 in the other two node_repair tests (_nval and _big).

src/riak_kv_node_cli.erl
```diff
@@ -0,0 +1,248 @@
...
+    ].
+
+node_repair_start_usage() ->
+    ["riak admin node repair start [-n|--node NODE]\n",
+     "Start partition repair on NODE.\n"
+    ].
+
+node_repair_stop_usage() ->
+    ["riak admin node repair stop [-n|--node NODE|all] REASON\n",
+     "Kill all an ongoing partition repair on NODE, with REASON.\n"
```
martinsumner: "Kill all ongoing"

src/riak_kv_node_cli.erl
```diff
@@ -0,0 +1,248 @@
...
+            Table =
+                [begin
+                     Rows =
+                         [[{mod, Mod}, {idx, integer_to_binary(Idx)}, {pid, list_to_binary(pid_to_list(Pid))}]
+                          || {Mod, Idx, Pid} <- NRes],
+                     case Rows of
+                         [] ->
+                             text("No active node repairs on ~s\n", [Node]);
+                         _ ->
+                             [text("Node repairs on ~s", [Node]), table(Rows)]
```
martinsumner: This is a bit confusing, as what is being reported is a list of vnode repairs, not a list of node repairs.

Would this be better? -

"Vnode repairs triggered by node repair on ~s"

src/riak_kv_node_cli.erl
```diff
@@ -0,0 +1,248 @@
...
+    end.
+
+get_node_repair_status(Nodes) ->
+    [begin
+         Vnodes = erpc:call(Node, riak_core_vnode_manager, all_vnodes, []),
+         Statuses = [{Mod, Idx, Pid,
+                      erpc:call(Node, riak_core_vnode_manager, repair_status, [{Mod, Idx}])}
+                     || {Mod, Idx, Pid} <- Vnodes],
+         {Node, [{Mod, Idx, Pid} || {Mod, Idx, Pid, Status} <- Statuses, Status /= not_found]}
+     end || Node <- Nodes].
```
martinsumner, line 147: What is the purpose of reporting the pid here?  We should at least be clear that this is the vnode Pid , so no-one assumes that the pid is associated with the transfer, and kills the pid not realising they're killing the vnode.

src/riak_kv_node_cli.erl
hmmr: Changed the printed message accordingly.

src/riak_kv_node_cli.erl
hmmr: `pid` dropped from the table.

src/riak_kv_node_cli.erl
hmmr: Corrected.

martinsumner (approved):

PR #158, opened 2026-06-16
When creating a (fake) new actor epoch entry - keep LMD unmodified

martinsumner:
A new actor epoch change is only made on an uncoordinated PUT, so it does not alter the object Last-Modified-Date - and likewise it should not alter the assumed LMD from the vector clock.

Looking at the comments on riak_kv_vnode:maybe_new_actor_epoch/2 - this update is only so that a new "frontier event" is created the next time this object is coordinated on the vnode.  The timestamp is irrelevant:

https://github.com/OpenRiak/riak_kv/blob/abcfdb761fa113348429281b59e97038b5b2dcce/src/riak_kv_vnode.erl#L4777-L4786


martinsumner:
https://github.com/basho/riak_kv/issues/679

martinsumner:

martinsumner:
Reviewed through discussion on Slack - @aef- and @hmmr

PR #160, opened 2026-06-17
Improve handling of large deltas

martinsumner:
See https://github.com/OpenRiak/riak_kv/issues/151

There is a need for an improved toolset in-built into Riak to handle large deltas, including those related to tombstone differences where the recovery could be impacted by key amnesia.

The changes are:

- A new internal bucket-sync function, that will allow for a bucket to be sync'd by slices of the segment space.  Testing in riak_test has o(50K) deltas in o(1m) objects being fixe d in o(100) seconds.
  - This bucket sync function uses new options exposed in https://github.com/OpenRiak/kv_index_tictactree/pull/13 to improve speed;
  - Allow for future configurability as well as a simple helper function;
- Add a hidden run-time configuration option to disable new actor key amnesia protection;
- Reduce the batch sizes in erase_keys/reap_tombs folds.

Tested in https://github.com/OpenRiak/riak_test/pull/30

### DIFF
--- a/src/riak_kv_ttaaefs_manager.erl
+++ b/src/riak_kv_ttaaefs_manager.erl
@@ -57,7 +57,6 @@
         disable_tree_repairs/0,
         disable_tree_reduction/0,
         enable_tree_reduction/0,
-        resync_bucket/1,
         resync_bucket/6
     ]
 ).
@@ -234,12 +233,6 @@ disable_tree_reduction() ->
 enable_tree_reduction() ->
     application:unset_env(riak_kv, ttaaefs_reduction).
 
-%% @doc
-%% Resync a bucket with some sane defaults.
--spec resync_bucket(riak_object:bucket()) -> ok.
-resync_bucket(Bucket) ->
-    resync_bucket(Bucket, all, all,128, 60 * 60 * 1000, 4).
-
 -spec get_current_scope() -> sync_config().
 get_current_scope() ->
     gen_server:call(?MODULE, get_current_scope, infinity).
@@ -278,6 +271,11 @@ resync_bucket(Bucket, KeyRange, DateRange, Width, Timeout, Loops) ->
             ?EXCHANGE_PAUSE_MS
         ),
     application:set_env(riak_kv, tictacaae_exchangepause, 1),
+    ?LOG_INFO(
+        "Attempt to resync bucket stared with ~w loops and ~w segment ranges",
+        [Loops, length(ShuffledSegRangeList)]
+    ),
+    SW = os:system_time(second),
     loop_resync(
         Bucket,
         KeyRange,
@@ -285,6 +283,10 @@ resync_bucket(Bucket, KeyRange, DateRange, Width, Timeout, Loops) ->
         ShuffledSegRangeList,
         Timeout,
         Loops
+    ),
+    ?LOG_INFO(
+        "Resync attempt completed in duration=~w seconds",
+        [os:system_time(second) - SW]
     ),
     enable_tree_reduction(),
     application:set_env(riak_kv, tictacaae_exchangepause, CurrentPause),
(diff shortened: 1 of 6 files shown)


src/riak_kv_vnode.erl
```diff
@@ -252,9 +252,18 @@
...
     %% Parallel AAE store rebuilds
 
-
--define(REAPER_BATCH_SIZE, 1024).
--define(ERASER_BATCH_SIZE, 1024).
+%% Larger batches are potentially more efficient, but as RingSize increases, a
+%% higher batch size will lead to more results being queued on the node
+%% coordinating the query rather than being distributed.
+%% 
+%% e.g. with RS = 1024 than a result set of < 40K overall would all be queued
```
ThomasArts:
```suggestion
%% e.g. with RS = 1024 a result set of < 40K overall would all be queued
```

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -204,6 +215,91 @@ enable_ssl(Enable, Credentials) ->
...
+%% @doc
+%% When performing per-bucket merges, the reduction check can be disabled with
+%% this function.  If the range being covered is static (i.e. no mutating 
+%% objects are covered, a needless verification of merge_tree_range can be
+%% avoided with this option).
+%% 
+%% This is a runtime change only, it cannot be permanently enabled across
+%% restarts.
+-spec disable_tree_reduction() -> ok.
+disable_tree_reduction() ->
```
ThomasArts, line 224: I guess there are good reasons to use this "global variable" in the application environment, but in theory a local state variable had been cleaner.

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -713,36 +852,159 @@ code_change(_OldVsn, State, _Extra) ->
...
+    WRF =
+        case application:get_env(riak_kv, ttaaefs_reduction) of
+            {ok, Scale} when is_float(Scale), Scale >= 0.0, Scale =< 1.0 ->
+                [{worthwhile_reduction, Scale}];
+            _ ->
+                []
+        end,
+    WRC =
+        case application:get_env(riak_kv, ttaaefs_reduction_cached) of
+            {ok, Count} when is_integer(Count), Count >= 0 ->
```
ThomasArts, line 999: This was enforced >0 when setting it as option, now >= 0.

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -814,6 +1076,94 @@ maybe_repair_trees(LastRepairID, _Filtered) ->
...
+        fun(I) -> 
+            lists:seq(I, I + HalfW - 1)
+            ++ lists:seq(I + 16#8000, I + 16#8000 + HalfW - 1)
+        end,
+        lists:seq(0, 16#7FFF, HalfW)
+    ).
+
+%% @doc
+%% Shuffle a list, inefficiently and inaccuratly.  Can do better when shuffle
+%% function is introduced and OTP.  The inefficiency and inaccuracy are
```
ThomasArts:
```suggestion
%% function is introduced in OTP.  The inefficiency and inaccuracy are
```

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -204,6 +215,91 @@ enable_ssl(Enable, Credentials) ->
...
+    pos_integer(),
+        % Timeout for an individual sync attempt within the loop
+    pos_integer()
+        % how many times to loop, to complete sync if all segments are
+        % mismatched will need to be at least:
+        % Width div (MaxResults * RangeBoost)
+) ->
+    ok.
+resync_bucket(Bucket, KeyRange, DateRange, Width, Timeout, Loops) ->
+    ShuffledSegRangeList = shuffle(generate_seg_lists(Width)),
```
ThomasArts, line 259: This can be rather expensive indeed. These lists are potentailly large and you go over them 3 times.
But there is a remark that this need not be efficient, so most likely fine.
If you were to do it more efficient, then generation and shuffling as one operation would make sense.

ThomasArts (approved):

src/riak_kv_ttaaefs_manager.erl
martinsumner: See https://github.com/OpenRiak/kv_index_tictactree/pull/13#discussion_r3451277659 - now >=0

src/riak_kv_ttaaefs_manager.erl
martinsumner: This happens once, on a single process, at the start of a batch of work that will last several minutes prompting work across the cluster.  Taking a millisecond or so extra is fine.

src/riak_kv_ttaaefs_manager.erl
martinsumner: Comment removed, with clarifying note in commit.  I'm not sure what the correct decision is here, so keeping as-is for now.

Issue #165, opened 2026-07-01
Handle a crash of a worker pool when rebuilding AAE trees


martinsumner:
The workers running rebuilds (e.g. rebuild trees in a native AAE store) may crash.  This results in the PrematureAcc error being sent in the FinishFun:

https://github.com/OpenRiak/riak_kv/blob/48285d5d7aa72c4167e338b3d28ae9ba434cfcb7/src/riak_kv_worker.erl#L43-L52

The FinishFun sent to the fold does not handle anything other than `ok` - so what happens on PrematureAcc?

However, work may be queued prior to processing, and the worker pool could crash without the `handle_work/3` being called.

In some cases (if the work is discarded from the queue during shutdown) a reply of `{error, vnode_shutdown}` is sent:

https://github.com/OpenRiak/riak_core/blob/2903cdc194fa4d538dd6f6da359e8465c11bb3c2/src/riak_core_worker_pool.erl#L151

But the process requesting work, is probably the vnode at this stage.  This message is not explicitly handled.  does this mean a vnode crash?  (this might not be the worst thing).

In a production environment, the worker pool supervisor was reset whilst a rebuild was ongoing.  The end result appeared to be that the aae_controller was stuck in a state expecting for the rebuild to complete - without it actually completing.  This may have eventually led to an OOM error, but there wasn't an obvious work around other than restarting the node.

This needs testing, with a change to at least crash (the controller) when a rebuild process terminates unexpectedly.

PR #166, opened 2026-07-03
Monitor queues when used for rebuilds

martinsumner:
if a queue is used for rebuild, it is important to detect the failure of the pool, and not leave the AAE system in limbo.

Pools are monitored during a rebuild, and AAE is restarted should the pool fail before the rebuild completes.

The state has been extended on riak_kv_vnode:

- tictacaae_rebuilding now has {TS, Ref} not just TS as the state during rebuild, whereby Ref is a reference of a monitor of the pool;
- there is a a list of {reference(), Fun/1} pairs, if a 'DOWN' message is received for reference(), then the riak_kv_vnode state is updated with Fun(State) (and in this case this will restart AAE and update the controller pid on State).

Changes made to riak_kv_tictacaae_cli:

- Export of function for use in testing;
- Longer pause on sys:get_state (as the aae_controller restart in parallel mode may take up to 10s as there is an outstanding snapshot on the parallel store for the rebuild).

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -25,6 +25,7 @@
 -behaviour(application).
 -export([start/2, prep_stop/1, stop/1]).
 -export([check_kv_health/1]).
+-export([get_worker_pools/0]).
 
 -include_lib("kernel/include/logger.hrl").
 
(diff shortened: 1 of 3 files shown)


martinsumner:
https://github.com/OpenRiak/riak_kv/issues/165

martinsumner:
Tested in https://github.com/OpenRiak/riak_test/pull/32

